PR #231, opened 2023-05-01
Use ansys-tools-path

koubaa:
Move this functionality to the common installation path library.  This depends on the changes in https://github.com/pyansys/ansys-tools-path/pull/27 which will need a new release once added.

I also want to integrate it for embedding, I will update this PR later with that.


germa89 (approved):
LGTM.

But don't forget to add the package to ``pyproject.tml`` or ``setup.py``.

PR #257, opened 2023-05-29
Standardizing data paths

raph-luc:
Exact same changes as in https://github.com/ansys/pymapdl/pull/2064

Original issue tracker https://github.com/ansys/pyfluent/issues/1588

Changes Windows data path to `%LocalAppData%\Ansys\ansys_mechanical_core`, same naming scheme as in https://github.com/ansys/pyfluent/pull/1615, https://github.com/ansys/pymapdl/pull/2064, https://github.com/ansys/pyprimemesh/pull/485, https://github.com/ansys/ansys-tools-path/pull/44, https://github.com/ansys/pysystem-coupling/pull/158

Changed `SETTINGS_DIR` to point to `USER_DATA_PATH` to make sure these two are always the same and avoid unnecessary `appdirs` use

Removed try-except block

Behavior not changed in this PR: data path/settings dir and examples folders are immediately created on `import ansys.mechnical.core`, might be worth changing to create folders only when/if they are actually used

### DIFF
--- a/src/ansys/mechanical/core/__init__.py
+++ b/src/ansys/mechanical/core/__init__.py
@@ -4,6 +4,15 @@ import os
 
 import appdirs
 
+# Setup data directory
+USER_DATA_PATH = appdirs.user_data_dir(appname="ansys_mechanical_core", appauthor="Ansys")
+if not os.path.exists(USER_DATA_PATH):
+    os.makedirs(USER_DATA_PATH)
+
+EXAMPLES_PATH = os.path.join(USER_DATA_PATH, "examples")
+if not os.path.exists(EXAMPLES_PATH):
+    os.makedirs(EXAMPLES_PATH)
+
 from ansys.mechanical.core.logging import Logger
 
 # Create logger for package level use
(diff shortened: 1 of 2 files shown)


src/ansys/mechanical/core/__init__.py
```diff
@@ -4,6 +4,15 @@
...
 
+# Setup data directory
+USER_DATA_PATH = appdirs.user_data_dir(appname="ansys_mechanical_core", appauthor="Ansys")
+if not os.path.exists(USER_DATA_PATH):
+    os.makedirs(USER_DATA_PATH)
+
+EXAMPLES_PATH = os.path.join(USER_DATA_PATH, "examples")
+if not os.path.exists(EXAMPLES_PATH):
+    os.makedirs(EXAMPLES_PATH)
+
```
RobPasMue, line 15: Why removing the try/except? Pinging @koubaa here

raph-luc: Good question. The main reason is I don't see a reason for it (German didn't see a reason for it in pyMAPDL either). I am assuming that folder creation is important for this module (e.g. settings_dir) and/or tests, and so hiding exceptions here and failing silently does not look like a good idea to me (I may be wrong obviously). 

If this application does not need the folders created (and failing to create folders doesn't matter, so that we could hide potential failure to create/check for folders with this try-except block), then why are we even creating these folders here to begin with, every time the module is imported?

No strong opinions either way, just my thoughts. Let me know and I can easily add the try-except block back if preferred.

RobPasMue: Okay, sounds good to me then! This was probably "taken" from PyMAPDL nonetheless. Let's leave it without the try/catch =)

RobPasMue (approved):

PR #270, opened 2023-06-16
Add version configuration for embedding tests

koubaa:
By default, pytest will run embedding tests using 2023R2

This allows a new command line argument to specify the version

pytest --ansys-version 241

to request a specific version.

In the future, we can mark specific tests as disabled for specific versions, once we add features that only work in newer versions. That's not done in this branch

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -91,13 +91,13 @@ def ensure_embedding() -> None:
         raise Exception("Cannot run embedded tests if Mechanical embedding is not installed")
 
 
-def start_embedding_app() -> datetime.timedelta:
+def start_embedding_app(version) -> datetime.timedelta:
     from ansys.mechanical.core import App
 
     global EMBEDDED_APP
     ensure_embedding()
     start = datetime.datetime.now()
-    EMBEDDED_APP = App(version=232)
+    EMBEDDED_APP = App(version=int(version))
     startup_time = (datetime.datetime.now() - start).total_seconds()
     num_cores = os.environ.get("NUM_CORES", None)
     if num_cores != None:
@@ -111,9 +111,9 @@ EMBEDDED_APP = None
 
 
 @pytest.fixture(scope="session")
-def embedded_app(request):
+def embedded_app(pytestconfig, request):
     global EMBEDDED_APP
-    startup_time = start_embedding_app()
+    startup_time = start_embedding_app(pytestconfig.getoption("ansys_version"))
     terminal_reporter = request.config.pluginmanager.getplugin("terminalreporter")
     if terminal_reporter is not None:
         terminal_reporter.write_line(f"\t{startup_time}\tStarting Mechanical")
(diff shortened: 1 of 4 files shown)


koubaa:
@RobPasMue I also fixed the changelog (lots of things were missed).

Maybe there's a way to configure `dependabot` to update the changelog automatically?  And add a PR policy to block PR unless the changelog is modified too?

RobPasMue:
Hi @koubaa! I recommend you use the https://github.com/ansys/pyansys-tools-versioning repository (and Python package) for handling Ansys product versions within your library. Don't know if you have considered using it. PyMAPDL is using it right now.

And for the changelog policy adn automated tools, the team has worked planned on this area. Pinging @jorgepiloto since he was looking into how to include [towncrier](https://towncrier.readthedocs.io/en/stable/) into our actions.

tests/conftest.py
```diff
@@ -292,3 +292,8 @@ def mechanical_pool():
 
     assert f"Mechanical pool with {instances_count} active instances" in str(pool)
     pool.exit(block=True)
+
+
+def pytest_addoption(parser):
+    # parser.addoption("--debugging", action="store_true")
```
RobPasMue, line 298:
```suggestion
```
Remove it or uncomment it, as you prefer. I'd avoid having commented out code.

tests/scripts/embedding_log_test.py
```diff
@@ -7,34 +7,34 @@
 from ansys.mechanical.core.embedding.logger import Configuration, Logger
 
 
-def log_before_initialize():
+def log_before_initialize(version):
     """Write a log without initializing the embedded instance."""
     Logger.error("message")
```
RobPasMue, line 12: Why are you passing the version here if it is not used? Do you plan on writing a log message with it?

RobPasMue (approved):
LGTM

tests/scripts/embedding_log_test.py
koubaa: @RobPasMue look at line 55, I treat all the functions in this module the same way, passing the version in.  I'd rather do this than complicate the CLI implementation.

tests/conftest.py
koubaa: @RobPasMue I have the same preference, but I do intend to revisit this soon

koubaa:
@RobPasMue @germa89 

what is the relationship between https://github.com/ansys/ansys-tools-path
and https://github.com/ansys/pyansys-tools-versioning?

germa89:
> @RobPasMue @germa89
> 
> what is the relationship between https://github.com/ansys/ansys-tools-path and https://github.com/ansys/pyansys-tools-versioning?

In theory none, one just find the path of the installed product (`ansys-tools-path`) and the other (`ansys-tools-versioning`) is just to handle versions requirements, checks, etc.  For example `versioning` provides the following functions `requires_version`, and `server_meets_version` to check the product versions matches (or not) the server version.

tests/scripts/embedding_log_test.py
RobPasMue: However, you are not using it for anything, I'm I wrong? Passing in the ``version`` argument is not neccessary here, IMO

RobPasMue: Sure, sounds fine!

Issue #279, opened 2023-06-20
Need an option to start the application using a private appdata folder


koubaa:
Multiple mechanical instances can sometimes concurrently access the appdata folder without any synchronization, which leads to crashes or other issues.

We need an option (opt-in or opt-out) to use a private appdata folder, to preclude these sorts of issues. The option should be available to both remote session launch and embedded instance initialization

PR #285, opened 2023-06-30
Added private AppData functionality to embedding

klmcadams:
Created temporary user directories to allow for multiple instances of embedding to run in parallel. By default, private_appdata is False, meaning it will not create temporary user profiles when running an embedded instance. Since the temporary user directories are not fully removed at exit due to lingering licensing processes, I thought it would make sense to default private_appdata to False unless people want to run parallelly. I will be creating another issue to address fully removing the temporary directories. 

This PR relates to #279.

### DIFF
--- a/src/ansys/mechanical/core/embedding/app.py
+++ b/src/ansys/mechanical/core/embedding/app.py
@@ -1,6 +1,7 @@
 """Main application class for embedded Mechanical."""
 import atexit
 import os
+import shutil
 
 from ansys.mechanical.core.embedding import initializer, runtime
 from ansys.mechanical.core.embedding.config import Configuration, configure
@@ -22,14 +23,30 @@ def _dispose_embedded_app(instances):  # pragma: nocover
         instance._dispose()
 
 
+def _cleanup_private_appdata(folder):
+    try:
+        shutil.rmtree(folder)
+    except:
+        for root, dirs, files in os.walk(folder):
+            try:
+                for dir in dirs:
+                    shutil.rmtree(os.path.join(root, dir))
+            except:
+                pass
+
+
 class App:
     """Mechanical embedding Application."""
 
-    def __init__(self, db_file=None, **kwargs):
+    def __init__(self, db_file=None, private_appdata=False, **kwargs):
         """Construct an instance of the mechanical Application.
 
         db_file is an optional path to a mechanical database file (.mechdat or .mechdb)
         you may set a version number with the `version` keyword argument.
+
+        private_appdata is an optional setting for a temporary AppData directory.
+        By default, private_appdata is False. This is beneficial for running parallel
+        instances of mechanical.
         """
         global INSTANCES
         from ansys.mechanical.core import BUILDING_GALLERY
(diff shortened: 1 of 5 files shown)


src/ansys/mechanical/core/embedding/app.py
```diff
@@ -22,14 +23,30 @@ def _dispose_embedded_app(instances):  # pragma: nocover
         instance._dispose()
 
 
+def _cleanup_private_appdata(folder):
+    try:
+        shutil.rmtree(folder)
```
koubaa, line 28:
```
def _cleanup_private_appdata(folder):
    shutil.rmtree(folder, ignore_errors=True)
```

src/ansys/mechanical/core/embedding/initializer.py
```diff
@@ -68,6 +119,32 @@ def _get_default_version() -> int:
...
+    newProfile.mkdirs()
+
+    newProfile.copy_profiles()
+
+    newProfile.set_env()
+
+    warnings.warn(
+        "Please be advised that using the private_appdata option allows you to run "
+        "multiple pymechanical instances in parallel, but there are leftover files "
+        r"in a temporary directory:   C:\Users\<username><pid>.",
```
koubaa, line 141: What about "there may be leftover files in {profileName}" ?

tests/scripts/run_embedded_app.py
```diff
@@ -1,5 +1,48 @@
...
+def launch_app(appdata_option):
+    """Launch embedded instance of app."""
+    if appdata_option == "True":
+        app = pymechanical.App(private_appdata=appdata_option)
+    else:
+        app = pymechanical.App()
+    return app
+
+
+def set_false(appdata_option):
```
koubaa, line 16: I'd rather rename to `set_showtriad(appdata_option, value)` and remove `reset_showtriad` ?

tests/scripts/run_embedded_app.py
```diff
@@ -1,5 +1,48 @@
...
+    """Launch embedded instance of app & set ShowTriad to False."""
+    app = launch_app(appdata_option)
+    app.ExtAPI.Graphics.ViewOptions.ShowTriad = False
+    app.close()
+
+
+def check_showtriad(appdata_option):
```
koubaa, line 23: rename to print_showtriad()

tests/embedding/test_app.py
```diff
@@ -86,3 +86,80 @@ def test_warning_message(test_env, rootdir):
...
+
+    embedded_py = os.path.join(rootdir, "tests", "scripts", "run_embedded_app.py")
+
+    # Set ShowTriad to False
+    set_showtriad = subprocess.Popen(
+        [test_env.python, embedded_py, "True", "Set"],
+        stdout=subprocess.PIPE,
+        env=test_env.env,
+    )
+    set_showtriad.wait()
```
koubaa, line 111: imo better to do `subprocess.check_call()` instead of `Popen()` followed by `wait()`

koubaa (approved):

tests/scripts/run_embedded_app.py
klmcadams: That makes sense, I like that better as well

tests/embedding/test_app.py
klmcadams: I ended up doing check_output for the calls that need the stdout, but I changed the others Popen wait() sections to use check_call instead

PR #308, opened 2023-07-27
addin configuration and tests

koubaa:
Add the addin configuration and unit tests.

Need to find a way to restrict unit test to version 241

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -125,6 +125,6 @@ markers = [
     "python_env: tests that check for an appropriate python environment",
     "remote_session_launch: tests that launch Mechanical and work with gRPC server inside it",
     "remote_session_connect: tests that connect to Mechanical and work with gRPC server inside it",
-    "version_dependent: tests that run if ansys-version is greater than or equal to 241",
+     "minimum_version(num): tests that run if ansys-version is greater than or equal to the minimum version provided",
 ]
 xfail_strict = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,7 @@ doc = [
     "pypandoc==1.11",
     "pytest-sphinx==0.5.0",
     "pythreejs==2.4.2",
-    "pyvista==0.40.1",
+    "pyvista==0.41.1",
     "sphinx-autobuild==2021.3.14",
     "sphinx-autodoc-typehints==1.23.0", # BLOCKED BY sphinx-design - Cannot upgrade to Sphinx 7 for now
     "sphinx-copybutton==0.5.2",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -125,5 +125,6 @@ markers = [
     "python_env: tests that check for an appropriate python environment",
     "remote_session_launch: tests that launch Mechanical and work with gRPC server inside it",
     "remote_session_connect: tests that connect to Mechanical and work with gRPC server inside it",
+    "version_dependent: tests that run if ansys-version is greater than or equal to 241",
 ]
 xfail_strict = true
--- a/src/ansys/mechanical/core/embedding/app.py
+++ b/src/ansys/mechanical/core/embedding/app.py
@@ -25,7 +25,7 @@ def _cleanup_private_appdata(folder):
     shutil.rmtree(folder, ignore_errors=True)
 
 
-def _start_application(configuration: Configuration, version, db_file) -> "App":
+def _start_application(configuration: AddinConfiguration, version, db_file) -> "App":
     import clr
 
     clr.AddReference("Ansys.Mechanical.Embedding")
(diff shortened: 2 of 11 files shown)


koubaa:
@klmcadams this needs to wait until the backend change goes in, and then the docker container is updated.

Do you know how to restrict a test case to a specific version?

klmcadams:
> @klmcadams this needs to wait until the backend change goes in, and then the docker container is updated.
> 
> Do you know how to restrict a test case to a specific version?

I'm not sure, but I can look into it. Which unit test(s) do you only want to run 241?

klmcadams (approved):

tests/embedding/test_logger.py
```diff
@@ -72,6 +72,18 @@ def test_logging_write_info_after_initialize_with_error_level(rootdir, pytestcon
     assert "0xdeadbeef" not in stderr
 
 
+@pytest.mark.parametrize("addin_configuration", ["Mechanical", "WorkBench"])
+@pytest.mark.embedding
+def test_addin_configuration(rootdir, pytestconfig, addin_configuration):
```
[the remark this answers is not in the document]
RobPasMue: Is there a way for you to know (through env variables for example) which version you are running on? If so... I sugggest you use ``pytest.mark.skipif()`` - see https://docs.pytest.org/en/7.1.x/how-to/skipping.html#id1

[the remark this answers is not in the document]
klmcadams: @koubaa @RobPasMue I added the version check by creating the pytest.mark.version_dependent marker to add to tests that require a version greater than or equal to 241

src/ansys/mechanical/core/embedding/app.py
```diff
@@ -27,6 +27,23 @@ def _cleanup_private_appdata(folder):
...
+
+    if configuration.no_act_addins:
+        os.environ["ANSYS_MECHANICAL_STANDALONE_NO_ACT_EXTENSIONS"] = "1"
+
+    addin_configuration_name = configuration.addin_configuration
+    # Starting with version 241 we can pass a configuration name to the constructor
+    # of Application
+    if version >= 241:
+        return Ansys.Mechanical.Embedding.Application(db_file, addin_configuration_name)
+    return Ansys.Mechanical.Embedding.Application(db_file)
```
RobPasMue, line 44:
```suggestion
    if version >= 241:
        return Ansys.Mechanical.Embedding.Application(db_file, addin_configuration_name)
    else:
        return Ansys.Mechanical.Embedding.Application(db_file)
```

RobPasMue: Provides more clarity on how the source code behaves - even though the effective behaviour is the same, ``if...else`` is clearer when reading code (especially for these circumstances in which it is either one API or the other.

RobPasMue (approved):

pyproject.toml
```diff
@@ -125,5 +125,6 @@ markers = [
     "python_env: tests that check for an appropriate python environment",
     "remote_session_launch: tests that launch Mechanical and work with gRPC server inside it",
     "remote_session_connect: tests that connect to Mechanical and work with gRPC server inside it",
+    "version_dependent: tests that run if ansys-version is greater than or equal to 241",
```
koubaa, line 128: not quite - we will use this marker for other future versions too

tests/embedding/test_logger.py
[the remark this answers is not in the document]
koubaa: @klmcadams @RobPasMue see my other comment - we should make this marker accept a parameter.

`@pytest.mark.minimum_version(241)`

tests/scripts/embedding_log_test.py
```diff
@@ -4,6 +4,7 @@
 import sys
 
 import ansys.mechanical.core as mech
+from ansys.mechanical.core.embedding.config import Configuration as EmbeddingConfiguration
```
koubaa, line 7: @RobPasMue @klmcadams I'm not too happy about two classes that are both called Configuration.  Any suggestions for renaming them?

RobPasMue: I'd try to avoid classes with the same nomenclature inside the same library. Better go with ``EmbeddingConfiguration`` and ``RemoteConifguration``.

koubaa: @RobPasMue ok, we can do something like that.  I think the names should be `LoggingConfiguration` and `AddinConfiguration` since its not related to embedding/remote.  I don't know why I aliased it as `EmbeddingConfiguration` here.

But this is a breaking change (esp for logging configuration, which might already be used).  How should we do that?

koubaa: @RobPasMue client side!

koubaa: @RobPasMue I changed the name of the addin configuration class without any deprecation and created an issue for the logging configuration

PR #311, opened 2023-07-31
Add dev ci_cd yaml file

klmcadams:
Allows user to automatically trigger dev_ci_cd workflow to test embedding & remote tests on whichever branch you choose. Default revn is 241.

### DIFF
--- a/.github/workflows/dev_ci_cd.yml
+++ b/.github/workflows/dev_ci_cd.yml
@@ -1,0 +1,330 @@
+name: Dev GitHub CI
+# on:
+#   pull_request:
+#   workflow_dispatch:
+#   push:
+#     tags:
+#       - "*"
+#     branches:
+#       - main
+#       - release/*
+
+env:
+  PYMECHANICAL_PORT: 10000  # default won't work on GitHub runners
+  PYMECHANICAL_START_INSTANCE: FALSE
+  DOCKER_PACKAGE: ghcr.io/ansys/mechanical
+  DOCKER_IMAGE_VERSION: 24.1.0
+  DOCKER_MECH_CONTAINER_NAME: mechanical
+  MAIN_PYTHON_VERSION: '3.10'
+  PACKAGE_NAME: ansys-mechanical-core
+  PACKAGE_NAMESPACE: ansys.mechanical.core
+  DOCUMENTATION_CNAME: mechanical.docs.pyansys.com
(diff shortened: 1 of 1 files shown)


.github/workflows/dev_ci_cd.yml
```diff
@@ -0,0 +1,330 @@
...
+          pip3 install --trusted-host pypi.org --trusted-host pypi.python.org --trusted-host files.pythonhosted.org pip setuptools
+          pip3 install --upgrade pip flit
+      - name: Install packages for testing
+        run: |
+          pip install .[tests]
+
+      - name: Unit Testing and coverage
+        env:
+          LICENSE_SERVER: ${{ secrets.LICENSE_SERVER }}
+          ANSYSCL${{ github.event.inputs.revn }}_DIR: /install/ansys_inc/v${{ github.event.inputs.revn }}/licensingclient
```
klmcadams, line 168: I feel like this wouldn't work, but it won't let me test the workflow in this branch.

.github/workflows/dev_ci_cd.yml
RobPasMue, line 1: Is this workflow the same as in ci_cd? What are the main differences?

Maintaining two very similar workflows typically leads to maintenance issues (i.e. you forget to update one or the other from time to time)

.github/workflows/dev_ci_cd.yml
RobPasMue: That's not gonna work indeed... You should output the variable to GITHUB_ENV if you want to create "custom" named variable. See https://github.com/ansys/actions/blob/main/build-wheelhouse/action.yml#L87 for example

.github/workflows/dev_ci_cd.yml
koubaa: @klmcadams I was thinking the branch would just point the usual ci/cd to the new backend container version.  Is there a reason why this should be a separate yml?

klmcadams: @koubaa No, that makes sense! I can make these changes to the main yaml file instead on this branch. @RobPasMue I thought we only wanted to run the tests for 241, so that's why I made a separate yaml file containing only tests.

RobPasMue: Great, let's try reusing the existing workflow then =)

PR #319, opened 2023-08-01
Added warning for ansys-mechanical when provided an input script

klmcadams:
Added user warning when ansys-mechanical is run with the -i option (input script).

### DIFF
--- a/src/ansys/mechanical/core/run.py
+++ b/src/ansys/mechanical/core/run.py
@@ -4,6 +4,7 @@ import asyncio
 from asyncio.subprocess import PIPE
 import os
 import sys
+import warnings
 
 import ansys.tools.path as atp
 import click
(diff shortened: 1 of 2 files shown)


RobPasMue (approved):

src/ansys/mechanical/core/run.py
```diff
@@ -170,6 +171,11 @@ def cli(
     if input_script:
         args.append("-script")
         args.append(input_script)
+        warnings.warn(
```
koubaa, line 174: @klmcadams as of last night's build, this was fixed for 2024 R1.  So we should have a version check before this warning

klmcadams: @koubaa This is being addressed in #323

PR #320, opened 2023-08-01
Remove project lock file on close

### LINKED ISSUE #246: remove lock file when the Application object is cleaned up
When opening a mechdb file in pymechanical, at the end of the script, the lock file still exists on disk. This prevents opening the mechdb again without manually deleting the lock file.

When the application exits, the lock file should be removed.  This might be possible to support simply by adding an app.close() to the atexit handler.

klmcadams:
Added code to remove the project's lock file when the application closes & created a unit test to ensure the lock file is being removed. This closes #246.

### DIFF
--- a/src/ansys/mechanical/core/embedding/app.py
+++ b/src/ansys/mechanical/core/embedding/app.py
@@ -115,7 +115,13 @@ class App:
         self.DataModel.Project.New()
 
     def close(self):
-        """Close the application."""
+        """Close the application and remove the lock file."""
+        lockfile_path = os.path.join(self.DataModel.Project.ProjectDirectory, ".mech_lock")
+        if os.path.exists(lockfile_path):
+            try:
+                os.remove(lockfile_path)
+            except:
+                raise Exception("There was an issue removing the lock file.")
         self.ExtAPI.Application.Close()
 
     def execute_script(self, script: str):
--- a/src/ansys/mechanical/core/embedding/app.py
+++ b/src/ansys/mechanical/core/embedding/app.py
@@ -115,14 +115,14 @@ class App:
         self.DataModel.Project.New()
 
     def close(self):
-        """Close the application and remove the lock file."""
+        """Close the active project."""
+        # Call New() to remove the lock file of the
+        # current project on close.
+        self.DataModel.Project.New()
+
+    def exit(self):
+        """Exit the application."""
         self.ExtAPI.Application.Close()
-        lockfile_path = os.path.join(self.DataModel.Project.ProjectDirectory, ".mech_lock")
-        if os.path.exists(lockfile_path):
-            try:
-                os.remove(lockfile_path)
-            except:
-                raise Exception("There was an issue removing the lock file.")
 
     def execute_script(self, script: str):
         """Execute the given script with the internal IronPython engine."""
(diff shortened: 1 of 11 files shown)


src/ansys/mechanical/core/embedding/app.py
```diff
@@ -115,7 +115,13 @@ def new(self):
...
     def close(self):
-        """Close the application."""
+        """Close the application and remove the lock file."""
+        lockfile_path = os.path.join(self.DataModel.Project.ProjectDirectory, ".mech_lock")
+        if os.path.exists(lockfile_path):
+            try:
+                os.remove(lockfile_path)
+            except:
+                raise Exception("There was an issue removing the lock file.")
         self.ExtAPI.Application.Close()
```
RobPasMue, line 125: What about this close operation? If an error is raised, do you still want to call close? Seems to me that if now an error is encountered when deleting the lock file (if it exists)... we won't be able to ever close the app.

I would consider adding this statement right before the "raise" as well, inside the "except"

klmcadams: Yes, we should close the app either way. Thanks!

koubaa: @klmcadams I think we should remove the lock file _after_ closing, but ideally Close() itself should remove the lock file.  Can you start an email thread with Lei Gong about this?

src/ansys/mechanical/core/embedding/app.py
```diff
@@ -116,13 +116,13 @@ def new(self):
 
     def close(self):
         """Close the application and remove the lock file."""
+        self.ExtAPI.Application.Close()
```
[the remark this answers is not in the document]
koubaa: @RobPasMue I think if it fails I'd rather the exception not get caught, since `close` doesn't know how the user wants to handle the error.  imo exceptions should only get caught in libraries if the library knows how to handle it.

src/ansys/mechanical/core/embedding/app.py
```diff
@@ -115,8 +115,14 @@ def new(self):
         self.DataModel.Project.New()
 
     def close(self):
-        """Close the application."""
+        """Close the application and remove the lock file."""
```
koubaa, line 118: change the docstring to:

"""Close the active project."""

Let's add a new function `exit` for closing the application, in which we can internally call `Application.Close`, I will also research in the backend what we have to do in `exit` in order to properly terminate everything

tests/embedding/test_app.py
```diff
@@ -142,3 +142,19 @@ def test_normal_appdata(test_env, rootdir):
 
     # Assert ShowTriad was set to False for regular embedded session
     assert stdout == "False"
+
+
+@pytest.mark.embedding
+def test_rm_lockfile(embedded_app):
+    """Test lock file is removed on close of embedded application."""
+    mechdat_path = os.path.join(os.getcwd(), "test.mechdat")
```
koubaa, line 150: use a temporary folder, not the cwd

src/ansys/mechanical/core/embedding/app.py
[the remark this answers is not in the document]
RobPasMue: Okay, no try-except block in that case =)

koubaa (approved):

PR #323, opened 2023-08-03
Added version check for ansys-mechanical warning message

klmcadams:
Only show warning message for input script if revision is < 241.

### DIFF
--- a/src/ansys/mechanical/core/run.py
+++ b/src/ansys/mechanical/core/run.py
@@ -171,11 +171,13 @@ def cli(
     if input_script:
         args.append("-script")
         args.append(input_script)
-        warnings.warn(
-            "Please ensure ExtAPI.Application.Close() is at the end of your script. "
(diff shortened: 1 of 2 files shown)


RobPasMue (approved):

src/ansys/mechanical/core/run.py
```diff
@@ -177,6 +177,13 @@ def cli(
             stacklevel=2,
         )
 
+        if version < 241:
+            warnings.warn(
```
koubaa, line 181: @klmcadams looks like the warning is duplicated, and I  think warning only makes sense in batch mode

klmcadams: Ohh okay, I thought we wanted it with -script. I'll change it for batch mode only

koubaa: @klmcadams yep, its for the combination of -i and -b

PR #332, opened 2023-08-04
Added TempPathFactory to test_app_save_open

### LINKED ISSUE #330: use TempPathFactory in test_app_save_open

klmcadams:
Added TempPathFactory to test_app_save_open to remove need for manually deleting project files in an exception. Closes #330


tests/embedding/test_app.py
```diff
@@ -28,23 +27,17 @@ def test_app_save_open(embedded_app):
...
-        embedded_app.save()
-        embedded_app.new()
-        embedded_app.open(project_file)
-        assert embedded_app.DataModel.Project.Name == "PROJECT 2"
-        embedded_app.new()
-    except:
-        # clean up the project files
-        shutil.rmtree(project_files)
-        os.remove(project_file)
+    embedded_app.new()
```
klmcadams, line 32: Should I still keep the try except, but have "pass" in except instead of the folder & file removal?

koubaa: @klmcadams the `except` block was only necessary to not leak files during test runs, its not needed anymore

koubaa (approved):

PR #340, opened 2023-08-08
Update CHANGELOG.md

klmcadams:


koubaa (approved):
approving... but the clr-loader dependency fix needs to be released, since new users of pymechanical would encounter the issue when installing

klmcadams:
> approving... but the clr-loader dependency fix needs to be released, since new users of pymechanical would encounter the issue when installing

Thanks! I just had to move the updates we made to the unreleased section since they were in yesterday's release section. Whenever I do the patch release, I'll change the unreleased section to today's date instead.

PR #341, opened 2023-08-08
Max parallel 2 for embedding tests - ci_cd.yml

klmcadams:
Added "max-parallel: 2" for embedding test runs, since the license server should be able to handle two concurrent runs. This could resolve manually cancelling and re-running embedding tests that are running longer than expected.


RobPasMue:
You might need to update the changelog here 😄

RobPasMue (approved):

RobPasMue (approved):

PR #354, opened 2023-08-17
dependabot changelog automation

klmcadams:
Automatically creates a changelog for dependabot PRs

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -93,6 +93,23 @@ jobs:
               exit 1
           fi
 
+  dependabot-changelog:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: dangoslen/dependabot-changelog-helper@v3
+        with:
+          version: 'Unreleased'
+          activationLabel: 'dependabot'
+          changelogPath: './CHANGELOG.md'
+
+      - uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: "Updated Changelog"
+
   tests:
     name: Testing and coverage - Mechanical ${{ matrix.mechanical-version }}
     runs-on: public-ubuntu-latest-8-cores


.github/workflows/ci_cd.yml
```diff
@@ -93,6 +93,23 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: dangoslen/dependabot-changelog-helper@v3
+        with:
+          version: 'Unreleased'
+          activationLabel: 'dependabot'
```
RobPasMue, line 106: Your dependabot PRs will have to add this label now

.github/workflows/ci_cd.yml
```diff
@@ -93,6 +93,23 @@ jobs:
               exit 1
           fi
 
+  dependabot-changelog:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
RobPasMue, line 101: The token should have write permissions, otherwise it will fail

.github/workflows/ci_cd.yml
```diff
@@ -93,6 +93,23 @@ jobs:
...
+
+      - uses: dangoslen/dependabot-changelog-helper@v3
+        with:
+          version: 'Unreleased'
```
RobPasMue, line 105: Also, you might not even need to define the version since you are in line with their default

klmcadams: Yeah, I was wondering the same thing. I can remove it

.github/dependabot.yml
```diff
@@ -15,3 +16,4 @@ updates:
     labels:
       - "maintenance"
       - "dependencies"
+      - "dependabot"
```
RobPasMue, line 19: This label will also have to be created in the ``labels.yml``

.github/workflows/ci_cd.yml
RobPasMue: Hm you're right. They changed this at some point I guess. Previously it was only read permissions. Thanks @klmcadams

RobPasMue (approved):
LGTM!

.github/workflows/ci_cd.yml
```diff
@@ -93,6 +93,22 @@ jobs:
               exit 1
           fi
 
+  dependabot-changelog:
+    runs-on: ubuntu-latest
```
RobPasMue, line 97:
```suggestion
  dependabot-changelog:
    name: Automate dependabot changelog
    runs-on: ubuntu-latest
```

PR #385, opened 2023-09-11
Set up daily run for 241 testing and added manual inputs for workflow dispatch

### LINKED ISSUE #363: Add daily runs of 241 testing

Run feat/dev-ci_cd on a nightly basis

Set up nightly trigger in workflow?

_No response_

klmcadams:
Added scheduled cron to run the tests in ci_cd.yaml using 241 only. 
- Note: the scheduled cron only runs on the main branch. Depending on how busy the github runners are, there may be a 5-15 minute delay in the workflow actually running. If it's delayed for too long, it will skip the run altogether. We may need to troubleshoot a different time if it ends up skipping the scheduled run each day. 
- The cron is in UTC time, so it runs at 5pm CT, 6pm EST, and midnight CEST.

Added inputs to the workflow_dispatch to give flexibility in testing code for different revisions. Currently, the default revision is 231, and the test revision is 232. The test revision is 232 because it was listed in the "experimental" section of the tests, where tests under that revision are allowed to fail. 

This closes #363.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -68,13 +68,14 @@ jobs:
             echo "$license_name=$revn" >> $GITHUB_OUTPUT
           }
 
-          if [[ -z "${{ inputs.default-revn }}" && -z "${{ inputs.test-revn }}" ]]; then
+          if [[ -z "${{ inputs.default-revn }}" || -z "${{ inputs.test-revn }}" ]]; then
             if ${{ github.event_name == 'schedule' }}; then
               parse_revns "${{ env.DEV_REVN }}" "default" "default_lic"
               parse_revns "${{ env.DEV_REVN }}" "test" "test_lic"
             else
               parse_revns "${{ env.DEFAULT_REVN }}" "default" "default_lic"
               parse_revns "${{ env.TEST_REVN }}" "test" "test_lic"
+            fi
           else
             parse_revns "${{ inputs.default-revn }}" "default" "default_lic"
             parse_revns "${{ inputs.test-revn }}" "test" "test_lic"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_cd.yml
```diff
@@ -1,7 +1,27 @@
...
   workflow_dispatch:
+    inputs:
+      default-revn:
+        type: choice
+        description: 'Default revision number for containers.'
+        options: 
+        - '241'
+        - '232'
+        - '231'
+        default: '231'  # ensure this is the same as env.DEFAULT_REVN
```
koubaa, line 16: @RobPasMue @klmcadams what do you think about running all versions?  Now that I think about it, we want to make sure PyMechanical version X supports all mechanical versions before it was released. I'm not sure if we do that level of testing for other pyansys libraries.

.github/workflows/ci_cd.yml
```diff
@@ -13,19 +33,61 @@ env:
...
+  revn-variations:
+    name: Save variations of revn
+    runs-on: ubuntu-latest
+    outputs:
+      default: ${{ steps.save-versions.outputs.default }}
+      default_lic: ${{ steps.save-versions.outputs.default_lic }}
+      test: ${{ steps.save-versions.outputs.test }}
+      test_lic:  ${{ steps.save-versions.outputs.test_lic }}
+    steps:
+      - id: save-versions
```
koubaa, line 59: why are licenses part of variations? are those version specific?

koubaa (approved):

.github/workflows/ci_cd.yml
RobPasMue: Some libraries do perform this kind of testing. I'd suggest that if possible, you do test against all of them

koubaa: @klmcadams @RobPasMue let's handle that in a separate issue without delaying this PR. For me right now the priority is to make sure 241 is covered before development of the backend closes

.github/workflows/ci_cd.yml
klmcadams: Yes, the license env variable is version specific: ANSYSCL232_DIR: /install/ansys_inc/v232/licensingclient, for example

PR #399, opened 2023-09-15
add nonblocking sleep

### LINKED ISSUE #263: Add a function for non blocking sleep

Request is to expose a function for putting the mechanical process in non-blocking sleep so that on-going actions can be completed during the wait time.

_No response_

_No response_

koubaa:
Fixes #263 

This could also be used for remote session - but I'm not sure how best to import the util file in that case (or if we should do something different there, since util might eventually contain things that aren't usable in the remote interface)


koubaa:
I didn't add any documentation yet... rather keep this undocumented for now

klmcadams (approved):

PR #409, opened 2023-09-21
Warn for obsolete apis

koubaa:
Requires changes to the backend in 241 that are not yet merged


dipinknair (approved):

src/ansys/mechanical/core/embedding/warnings.py
```diff
@@ -0,0 +1,85 @@
...
+
+# TODO: Investigate `warnings.showwarning = my_special_function` in order to better
+#       control the printing behavior of warnings. But this is global state, so the
+#       function should only affect warnings that are thrown here.
+
+
+def _get_method_to_check(method_info: typing.Any) -> typing.Any:
+    """Get the reflection method to check for obsolete attributes.
+
+    If its a getter or setter, it has to be done in a different way.
```
MaxJPRey, line 36:
```suggestion
    If it's a getter or setter, it has to be done in a different way.
```

MaxJPRey (approved):
LGTM.

src/ansys/mechanical/core/embedding/warnings.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    if not method_info.IsSpecialName:
+        return method_info
+    type_props = [prop for prop in method_info.DeclaringType.GetProperties()]
+    getters = [t for t in type_props if t.GetGetMethod() == method_info]
+    setters = [t for t in type_props if t.GetSetMethod() == method_info]
+    if len(getters) > 0:
+        return getters[0]
+    if len(setters) > 0:
+        return setters[0]
```
MaxJPRey, line 46: I am guessing here getters come first and then setters because we might not always have setters?

PR #413, opened 2023-09-22
Update examples page

klmcadams:
- Moved examples/00_basic and examples/01_tips_n_tricks from the pymechanical repo to the pymechanical-examples repo [See related PR](https://github.com/ansys/pymechanical-examples/pull/95)
- Added an example script that implements the same example in a remote session or an embedded instance
- Changed the layout of the examples landing page to only contain cards
    - Embedding Examples: links to the embedding examples URL
    - Remote Session Examples: links to the remote session examples URL
    - Embedding Instance and Remote Session Examples: links to the remote session & embedding instance script

examples_landing_page
embedding_remote_example


RobPasMue:
Have a look at the rendered version of my proposal @klmcadams - see last commit. You can revert it if you prefer some other layout.

koubaa (approved):

examples/embedding_n_remote/README.txt
```diff
@@ -0,0 +1,4 @@
+Embedding and Remote Session Example
+====================================
+
+Below is the same example first demonstrated using an embedding instance, and next demonstrated using a remote session.
```
PipKat, line 4:
```suggestion
Here is the same example, first demonstrated using an embedding instance and then demonstrated using a remote session.
```
Google dev doc style guide says not to use above, below rather preceding or following. Here, I don't think a locator is needed.

PR #433, opened 2023-10-03
Fixed appdata tests

klmcadams:
- Fixed normal_appdata and private_appdata tests
- Allow all embedding tests to run in workflow

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -306,7 +306,7 @@ jobs:
           ANSYS_WORKBENCH_LOGGING_FILTER_LEVEL: 2
           NUM_CORES: 1
         run: |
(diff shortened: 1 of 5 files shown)


koubaa (approved):
approving.. but I'm not sure I understand what was wrong with the appdata tests

klmcadams:
> approving.. but I'm not sure I understand what was wrong with the appdata tests

Whenever we would run the appdata tests in the container, they would fail because of the crash on shutdown in Linux, so it was returning a signal abort and failing them (I had previously used subprocess.check_call). I adjusted them to be similar to the tests/embedding/test_logger.py tests, where they look for a specific string in the stdout to make sure it's working correctly instead

koubaa:
@klmcadams oh! makes sense, thanks!

klmcadams:
> @klmcadams oh! makes sense, thanks!

No problem!

PR #439, opened 2023-10-04
pytest --ansys-version dependent on existing install

### LINKED ISSUE #435: pytest default for --ansys-version should be based on the what is installed
On a docker container where 241 but not 232 is installed, pytest tries to use 232. It should default to 232 only if 232 is installed, otherwise use the installed version.

Additionally, if it can't find an installed version, the tests should error out at the collection phase, rather than trying to run the tests.

klmcadams:
Make pytest fixture, --ansys-version, dependent on the existing install. If it can't find one, then it defaults to 232

Closes #435

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,7 +6,7 @@ import shutil
 import subprocess
 import sys
 
-import ansys.tools.path
+import ansys.tools.path as atp
 import pytest
 
 import ansys.mechanical.core as pymechanical
@@ -117,7 +117,7 @@ EMBEDDED_APP = None
 @pytest.fixture(scope="session")
 def embedded_app(pytestconfig, request):
     global EMBEDDED_APP
-    startup_time = start_embedding_app(pytestconfig.getoption("ansys_version"), pytestconfig)
+    startup_time = start_embedding_app(pytestconfig.getoption("--ansys-version"), pytestconfig)
     terminal_reporter = request.config.pluginmanager.getplugin("terminalreporter")
     if terminal_reporter is not None:
         terminal_reporter.write_line(f"\t{startup_time}\tStarting Mechanical")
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -299,8 +299,15 @@ def mechanical_pool():
 
 
 def pytest_addoption(parser):
+    mechanical_path = atp.get_mechanical_path(False)
```
koubaa, line 302: I wonder if it makes sense to run the tests at all if atp can't find the paths

tests/conftest.py
```diff
@@ -299,8 +299,15 @@ def mechanical_pool():
...
+
+    if mechanical_path == None:
+        parser.addoption("--ansys-version", default="232")
+    else:
+        mechanical_version = atp.version_from_path("mechanical", mechanical_path)
```
koubaa, line 307: What do we want it to do if there are multiple versions installed? just take the default from atp, or default to a specific version like 232 where the tests are known to be stable?

koubaa:
This definitely fixes the issue, and I am okay with merging it as is, but I left some comments about doubts that I have

tests/conftest.py
klmcadams: I think if atp can't find the path, then we could keep the default 232 option, but I see what you mean. If it can't find the path on a 241 machine, then it would be pointless to run 232 tests on it by default.

tests/conftest.py
klmcadams: Can we take the default from atp as long as it finds the mechanical_path above? I have multiple versions installed on my computer and atp defaulted to 232

RobPasMue: IMO, we should let atp decide - whatever the latest version is available, that's the one we take. atp provides them back in an orderly manner from latest to oldest, if I remember properly. That should be the behavior.

koubaa (approved):

PR #443, opened 2023-10-06
updated example-data url from raw to tree

Amrit-Hub:


Amrit-Hub:
raw/master is not working. The url link needs to be tree/master

klmcadams:
> raw/master is not working. The url link needs to be tree/master

Hi @Amrit-Hub, thanks for your message! We have to use "raw/master" rather than "tree/master" because of the download_file function:

```
from ansys.mechanical.core.examples.downloads import download_file

file = download_file("remote_script.py", "pymechanical", "embedding")
print(file) # location of where the file was downloaded
```

Whenever I replaced "raw" with "tree" in the url, it ended up downloading a file containing the html of the GitHub page where the file was located: https://github.com/ansys/example-data/tree/master/pymechanical/embedding/remote_script.py. Instead, we want to download the raw file contents: https://github.com/ansys/example-data/raw/master/pymechanical/embedding/remote_script.py. Let me know if you have any questions!

Issue #451, opened 2023-10-10
app.save does not work when trying to save already saved project in the current session


rajesh1359:
### 🔍 Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### 🐞 Description of the bug

app.save() with an empty input does not work as underlying mechanical API does not accept _None_ as the path.

### 📝 Steps to reproduce

app.save("D:/test.mechdat")

app.save()

### 💻 Which operating system are you using?

Windows

### 📀 Which ANSYS version are you using?

2024R1

### 🐍 Which Python version are you using?

3.8

### 📦 Installed packages

```shell
---
```

PR #453, opened 2023-10-10
Adds a check if path is None in app.save method

rajesh1359:
app.save() should be able to save already saved project as well. if we send None in DataModel.Project.Save(None) mechanical API then it throws error "Null or Empty File Path."

image

This is fix for issue: https://github.com/ansys/pymechanical/issues/451

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ This document follows the conventions laid out in [Keep a CHANGELOG](https://kee
  - Run all embedding tests & fix appdata tests ([#433](https://github.com/ansys/pymechanical/pull/433))
  - unset all logging environment variables ([#434](https://github.com/ansys/pymechanical/pull/434))
  - pytest --ansys-version dependent on existing install ([#439](https://github.com/ansys/pymechanical/pull/439))
- - Fix app.save method for saving already saved project in current session (https://github.com/ansys/pymechanical/pull/452)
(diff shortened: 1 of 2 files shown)


rajesh1359:
I am not sure how the test cases (https://github.com/ansys/pymechanical/actions/runs/6473844751/job/17577477416#step:11:55) are being pass successfully earlier as well.

With below snippet, I am able to reproduce.

`
from ansys.mechanical.core import embedding as app

from ansys.mechanical.core import global_variables

embedded_app = app.App(version=241)

globals().update(global_variables(embedded_app))

embedded_app.save("D:/test123.mechdat")
embedded_app.save()
`

rajesh1359:
This seems to be the behavior starting from 2024R1. Older versions take None as input to the DataModel.Project.Save(None). We can either revert the behavior in Mechanical API or we can make change in embedding source code to accommodate the new behavior.

koubaa:
@rajesh1359 I confirm that there was a change in 241. Passing a path to `ExtAPI.Project.Save()` is obsolete, and if the path passed in is null or empty, it throws an error. Since we already have a `save_as(path)` option on app, do you think we can remove the path argument from save (even in older releases)?

rajesh1359:
@koubaa Typically while saving the new project first time, user would use app.save(path) instead of app.save_as(path) (although mechanical API recommends using SaveAs). we should continue to take path input in app.save as well.

PR #454, opened 2023-10-10
env console with subprocess

dipinknair:


src/ansys/mechanical/core/run.py
```diff
@@ -241,7 +197,122 @@ def cli(
         #        the user only sees the message when the server is ready.
         print(f"Serving on port {port}")
 
-    _run(args, env)
```
koubaa, line 244: @dipinknair why change the ansys-mechanical script?

PR #461, opened 2023-10-15
Update obsolete api in quick workbench testing - embedding

dipinknair:


koubaa:
@dipinknair thanks for this! This API is also used in the internal QA tests where this came from. Could you please check to see if its still used there? (And send me a chat if it is)

koubaa (approved):

PR #466, opened 2023-10-16
Update temporary file creation in test_app

klmcadams:
use NamedTemporaryFile to create a temporary file instead of mktemp


RobPasMue (approved):
LGTM!

PR #506, opened 2023-11-22
Add cheetsheat template and improve example visibility

dipinknair:
image

Add cheat sheet on index page. and adjust the **examples** tab to 3rd position


RobPasMue:
Top bar looks completely off... get all titles to be in the same level

RobPasMue:
> Top bar looks completely off... get all titles to be in the same level

This is caused by the search bar mostly

dipinknair:
> > Top bar looks completely off... get all titles to be in the same level
> 
> This is caused by the search bar mostly

The search bar alignment issues comes from the fact that we have section title names which are long. By default we sphinx theme shows 5 sections and rest of them under **more**. To fix this,

- [ ] option 1
Reduce the number of sections shows in page to 4 and keep others under **more** like below.
___
image

___
or we can change the length of the title like below.  

- [ ] option 2
**_mechanical scripting -> Scripting
Remote Session -> Remote
Embedded Instance -> Embedded
API References -> API_**

@koubaa @RobPasMue . Which one do you think better?

RobPasMue:
I would go with option 1 - but up to you guys

dipinknair:
Final look @RobPasMue @koubaa 
image

klmcadams (approved):
Looks good to me!

RobPasMue:
Much better, thanks @dipinknair

PR #512, opened 2023-11-28
Add release note configuration

dipinknair:
Add release not config and more labels

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,16 @@
+changelog:
+  categories:
+    - title: New features or code improvements
+      labels:
+        - enhancement
+    - title: Bug fixes or behavior changes
+      labels:
+        - bug
+    - title: Documentation
+      labels:
+        - documentation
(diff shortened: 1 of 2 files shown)


.github/release.yml
```diff
@@ -0,0 +1,16 @@
+changelog:
```
RobPasMue, line 1: How does GitHub know to use this file? Is it standard procedure? Its the first time I see it

RobPasMue (changes requested):

.github/release.yml
dipinknair: if you see release notes. you can see the details from changelog which we need to manually generate. But with this it will be automatic. I am thinking in future we can incorporate release notes into the documentation. 
I was following the template from pymapdl. 

@germa89  let me know if my understanding is correct.

germa89: why me? XDD

RobPasMue (approved):

PR #516, opened 2023-12-04
Fix ansys-mechanical for finding path

### LINKED ISSUE #497: ansys-mechanical doesn't work on linux for 241
after running

`save-ansys-path --name mechanical /path/to/.workbench`

running ansys-mechanical console script
`ansys-mechanical -r 241 --exit -i test.py`

gives the following error:

ValueError: invalid literal for int() with base 10: ''

This is because we always use find_mechanical, which looks for mechanical in default locations. If the user saved the location using ansys.tools.path, it should use `get_mechanical_path()`

dipinknair:
Closes #497

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,19 +1,12 @@
 documentation:
-- doc/**/*
-- examples/**/*
+- changed-files:
+  - any-glob-to-any-file: ['doc/**/*', 'examples/**/*']
 maintenance:
-- .github/**/*
-- codecov.yml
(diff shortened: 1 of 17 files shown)


koubaa (approved):

dipinknair:
### Comparison

| Original | Modified |
|----------|----------|
| image |image|
| image |image|
---
----
Added ansys path saving under getting started

image

sbradrick:
No comments. Thanks.
________________________________

PR #541, opened 2023-12-22
ci: update ansys/actions to v5

jorgepiloto:
Fixes the issues in https://github.com/ansys/pymechanical/actions/runs/7300442441/job/19895293745. Dependabot is properly configured but updates on a weekly basis, so this update was no triggered yet.


dipinknair (approved):
Awesome ! Thanks for looking into it.

PR #590, opened 2024-02-06
Update workflow versions to run 241 and 242

klmcadams:
- Updated stable version to be 241 and test version to be 242
- Changed variable names in revn-variations (used "test" instead of experimental)
- Changed environment variable names (removed "experimental" env variable, only have "stable" or "dev")
- Added a config-matrix section to configure the matrix for tests depending on whether or not it's a scheduled run. If it's a scheduled run, it runs all of the mechanical versions in the tests. Otherwise, it only runs the tests with the test docker image version
 
The embedding test workflow instability will be addressed in a separate PR

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -2,26 +2,12 @@ name: GitHub CI
 
 on:
   pull_request:
+  workflow_dispatch:
   schedule:
     - cron: '00 22 * * *'  # UTC time, may start 5-15 mins later than scheduled time
-  workflow_dispatch:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_cd.yml
```diff
@@ -333,11 +300,21 @@ jobs:
           path: .coverage
           retention-days: 7
 
+      - name: Publish Test Report
+        uses: mikepenz/action-junit-report@v3
+        if: always()
+        with:
```
koubaa, line 306: Big improvement here!

PR #617, opened 2024-02-26
feat: add changelog action

klmcadams:
Add changelog action to pymechanical

This action will create md files containing the title of the PR. The md file will be named based on the PR number and what type of label is applied to the PR. For example, the file for this pull request will be 617.added.md and will be saved in the changelog.d folder. If you update the title of the PR, open the PR, or add a label (either manually or through the labeler.yml file), there will be a commit by the actions bot saying "adding fragment" (The message has now been changed to "Adding changelog entry: 617.added.md" in the action)

Once we are ready to do a release, run the following command to add all of the md files in changelog.d as entries to CHANGELOG.md under the specified version: ``towncrier build --yes --version 0.10.8``

**Note**: Do not include v in the version argument, or the url for the tag in the changelog will get messed up. The above command will remove all of the md files in changelog.d. If you want to keep the files, use the following command: ``towncrier build --keep --version 0.10.8``. 

If you want to see what the changelog will look like without making changes to CHANGELOG.md, use the following command: ``towncrier build --keep --draft --version 0.10.8``. This command will print out the changelog updates in the stdout


RobPasMue:
In that case, we should remove your current changelog action, right? The one that creates it automatically for dependabot and the one that checks that you are writing to the CHANGELOG:md file

klmcadams:
@RobPasMue that's right! I forgot about that

PR #621, opened 2024-02-26
Put remote example before embedding example

klmcadams:
Test if putting the remote example before the embedding example fixes the documentation issue in the workflow


examples/embedding_n_remote/embedding_remote.py
```diff
@@ -22,13 +22,130 @@
 
 """ .. _ref_embedding_remote:
 
-Embedding & Remote Example
+Remote & Embedding Example
 --------------------------
 This code, which uses the same example, first demonstrates how to use
-an embedding instance and then demonstrates how to use a remote session.
+a remote session and then demonstrates how to use an embedded instance.
```
klmcadams, line 28:
```suggestion
a remote session and then demonstrates how to use an embedding instance.
```

koubaa (approved):

samigithub2022 (approved):

PR #628, opened 2024-03-06
Fix : Update Release action to use Stable version of Mechanical

dipinknair:
Scheduled run :  242
PR : 241
Release : Use all stable versions (231,232,241)

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -152,6 +152,8 @@ jobs:
 
           echo "stable_container=${{ env.DOCKER_PACKAGE }}:${{ env.LATEST_STABLE_DOCKER_IMAGE_VERSION }}" >> $GITHUB_OUTPUT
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_cd.yml
```diff
@@ -152,6 +152,8 @@ jobs:
 
           echo "stable_container=${{ env.DOCKER_PACKAGE }}:${{ env.LATEST_STABLE_DOCKER_IMAGE_VERSION }}" >> $GITHUB_OUTPUT
 
+          cat $GITHUB_OUTPUT
```
RobPasMue, line 155: I'd avoid doing a cat of the whole GITHUB OUTPUT. Is this only for dev purposes ok your side @dipinknair ? If so, can we delete the cat commands?

dipinknair: Yes. removed !

koubaa (approved):

PR #636, opened 2024-03-11
Maint: Update nightly run image version

dipinknair:
Changing nightly run image version to take new candidate image (242)

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -134,8 +134,9 @@ jobs:
     steps:
       - id: save-versions
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_cd.yml
```diff
@@ -152,6 +152,8 @@ jobs:
 
           echo "stable_container=${{ env.DOCKER_PACKAGE }}:${{ env.LATEST_STABLE_DOCKER_IMAGE_VERSION }}" >> $GITHUB_OUTPUT
 
+          cat $GITHUB_OUTPUT
```
dipinknair, line 155: @RobPasMue  I feel this is convenient in case we want to see what image is being used. Since it does not contain any secret, do you think its okay?

RobPasMue: Fine

RobPasMue (approved):

PR #647, opened 2024-03-18
Feat: Block 32 bit python for embedding

dipinknair:
Closes: #145

### DIFF
--- a/src/ansys/mechanical/core/embedding/initializer.py
+++ b/src/ansys/mechanical/core/embedding/initializer.py
@@ -88,7 +88,7 @@ def _get_default_version() -> int:
     return int_version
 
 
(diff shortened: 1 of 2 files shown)


src/ansys/mechanical/core/embedding/initializer.py
```diff
@@ -88,8 +88,18 @@ def _get_default_version() -> int:
     return int_version
 
 
+def _check_python_interpreter_architecture():
```
RobPasMue, line 91: If you want it to be private and not used anywhere else apart from this module, I'd go with a double leading underscore.

koubaa (approved):

RobPasMue (approved):

PR #662, opened 2024-03-26
Add embedding_scripts marker

klmcadams:
Add embedding_scripts marker for tests that use subprocess to launch embedded apps


.github/workflows/ci_cd.yml
```diff
@@ -269,6 +269,80 @@ jobs:
...
+          LICENSE_SERVER: ${{ secrets.LICENSE_SERVER }}
+          ANSYSLMD_LICENSE_FILE: 1055@${{ secrets.LICENSE_SERVER }}
+          ANSYS_WORKBENCH_LOGGING_CONSOLE: 0
+          ANSYS_WORKBENCH_LOGGING: 0
+          ANSYS_WORKBENCH_LOGGING_FILTER_LEVEL: 2
+          NUM_CORES: 1
+          PYTHONUNBUFFERED: 1
+        run: |
+          . /env/bin/activate
+          xvfb-run mechanical-env pytest -m embedding_scripts -s --junitxml test_results_embedding_scripts${{ matrix.python-version }}.xml || true
```
koubaa, line 318: @klmcadams no need for the crash workarounds like `|| true` here. pytest will not crash because pymechanical embedding is only in the subprocess.

.github/workflows/ci_cd.yml
```diff
@@ -278,6 +278,80 @@ jobs:
           include_passed: true
           fail_on_failure: true
 
+  embedding-scripts-tests:
+    name: Embedding scripts testing and coverage
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    needs: [style, revn-variations]
```
dipinknair, line 285:
```suggestion
    needs: [smoke-tests, revn-variations]
```

dipinknair: I added this changes in last PR where we do smoke tests first which takes style. Then after smoke tests we do all embedding, remote tests.

.github/workflows/ci_cd.yml
```diff
@@ -278,6 +278,80 @@ jobs:
...
+          LICENSE_SERVER: ${{ secrets.LICENSE_SERVER }}
+          ANSYSLMD_LICENSE_FILE: 1055@${{ secrets.LICENSE_SERVER }}
+          ANSYS_WORKBENCH_LOGGING_CONSOLE: 0
+          ANSYS_WORKBENCH_LOGGING: 0
+          ANSYS_WORKBENCH_LOGGING_FILTER_LEVEL: 2
+          NUM_CORES: 1
+          PYTHONUNBUFFERED: 1
+        run: |
+          . /env/bin/activate
+          xvfb-run mechanical-env pytest -m embedding_scripts -s --junitxml test_results_embedding_scripts${{ matrix.python-version }}.xml
```
koubaa, line 327: minor - I don't think the embedding_scripts do any image export, so `xvfb-run` can be dropped

koubaa (approved):

.github/workflows/ci_cd.yml
klmcadams, line 327:
```suggestion
          mechanical-env pytest -m embedding_scripts -s --junitxml test_results_embedding_scripts${{ matrix.python-version }}.xml
```

PR #664, opened 2024-03-26
Test : pdf action 

dipinknair:
removes Setting num of cores in conf.py. This way pdf build not creating an embedded app and hence not giving error.


klmcadams (approved):

RobPasMue (approved):

koubaa:
@dipinknair I don't think this is correct. The default number of cores is 4. Does this mean that the doc build will not work on the default 2core ubuntu-latest???

dipinknair:
@koubaa  https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories
Ubuntu-latest is 4 core for public repos

PR #675, opened 2024-04-03
maint: use ansys-sphinx-theme "autoapi" target

klmcadams:
Use the ansys-sphinx-theme "autoapi" target.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,7 @@ tests = [
 ]
 doc = [
     "Sphinx==7.2.6",
-    "ansys-sphinx-theme==0.15.0",
+    "ansys-sphinx-theme[autoapi]==0.15.1",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -56,7 +56,7 @@ tests = [
 ]
 doc = [
     "Sphinx==7.2.6",
-    "ansys-sphinx-theme==0.15.0",
+    "ansys-sphinx-theme[autoapi]==0.15.1",
```
RobPasMue, line 59: I believe you are not using "autoapi" so it is not needed. Also, if you define this target you will eventually enter into conflicts with ``sphinx_design``. You should just install it through the target in that case (i.e. remove it from your deps)

RobPasMue (changes requested):

PR #680, opened 2024-04-03
first version of 3d visualization with pyvista

koubaa:


RobPasMue:
@koubaa - just FYI, we are working on a common viz interface to support multiple backends through the same API. I'll get back to you offline about this. It should be released fairly soon.

dipinknair (approved):

PR #688, opened 2024-04-08
Fix: Modify retrieving path of Mechanical in tests

dipinknair:
Use `get_mechanical_path` instead of `find_mechanical` in `remote_session_launch` tests

### DIFF
--- a/doc/changelog.d/688.fixed.md
+++ b/doc/changelog.d/688.fixed.md
@@ -1,0 +1,1 @@
+Fix: Modify retrieving path of Mechanical in tests
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -312,7 +312,7 @@ def mechanical_pool():
     if not pymechanical.mechanical.get_start_instance():
         return None
 
-    path, version = atp.find_mechanical()
+    path = atp.get_mechanical_path()
 
     exec_file = path
     instances_count = 2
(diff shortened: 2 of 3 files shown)


koubaa:
@dipinknair which issue is this associated with?

dipinknair:
> @dipinknair which issue is this associated with?

Came across this issue while debugging remote_session_launch test failures locally. Since find_mechanical() does not check config.txt , some test were failing because of not able to retrieve path.

koubaa:
@dipinknair we run on containers that put mechanical under /install/ansys_inc/vNnn.  So `find_mechanical` should work there. Do you know why it didn't?

dipinknair:
> @dipinknair we run on containers that put mechanical under /install/ansys_inc/vNnn. So `find_mechanical` should work there. Do you know why it didn't?

This is to facilitate testing locally so that we can debug further issues that we see in tests. This is not a fix for the test failures but  fix to run the test locally since Pool test are also using find_mechanical.

koubaa:
@dipinknair in that case I suggest leaving a fallback to find the path if there's nothing in the config, otherwise this will require setting the path before the test can run.

dipinknair:
> @dipinknair in that case I suggest leaving a fallback to find the path if there's nothing in the config, otherwise this will require setting the path before the test can run.

get_mechanical_path takes care that and `find` arg set always true for this function which falls back to find mechanical if not path found from the config file

PR #703, opened 2024-04-12
Test: add check for mechanical-env

dipinknair:
When not using mechanical-env for embedding in Linux , you should see this exception

For embedding tets:
image

For Docs build
image

### DIFF
--- a/src/ansys/mechanical/core/embedding/initializer.py
+++ b/src/ansys/mechanical/core/embedding/initializer.py
@@ -24,6 +24,7 @@
 from importlib.metadata import distribution
 import os
 from pathlib import Path
+import platform
 import sys
 import warnings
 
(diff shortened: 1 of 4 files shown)


src/ansys/mechanical/core/embedding/initializer.py
```diff
@@ -103,16 +104,28 @@ def _get_default_version() -> int:
...
     """Embedding support only 64 bit architecture."""
-    import platform
-
     if platform.architecture()[0] != "64bit":
         raise Exception("Mechanical Embedding requires a 64-bit Python environment.")
 
 
+def __check_for_mechanical_env():
+    """Embedding in linux platform must use mechanical-env."""
+    if platform.system() == "Linux" and os.environ.get("PYMECHANICAL_EMBEDDING") != "TRUE":
```
koubaa, line 113: looks good

PR #708, opened 2024-04-16
Feat: `mechanical-env` check before running embedding

dipinknair:
This check is to ensure that `mechanical-env` is used for embedding in linux.


src/ansys/mechanical/core/embedding/initializer.py
```diff
@@ -103,15 +104,25 @@ def _get_default_version() -> int:
...
-
     if platform.architecture()[0] != "64bit":
         raise Exception("Mechanical Embedding requires a 64-bit Python environment.")
 
 
+def __check_for_mechanical_env():
+    """Embedding in linux platform must use mechanical-env."""
+    if platform.system() == "Linux" and os.environ.get("PYMECHANICAL_EMBEDDING") != "TRUE":
+        raise Exception(
+            "Mechanical Embedding requires to use `mechanical-env` in Linux. "
```
koubaa, line 115: I suggest this wording:


On linux, embedding an instance of the Mechanical process using the App class requires running python
inside of a Mechanical environment. Use the `mechanical-env` script to do this. For more information, refer to:

koubaa (approved):

PR #709, opened 2024-04-16
ci: Add vulnerability check

dipinknair:
This failure of the action means mechanical has vulnerabilities which we should look into @ansys/pymechanical-maintainers


.github/workflows/ci_cd.yml
```diff
@@ -57,6 +57,18 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
```
dipinknair, line 57: use new version of ansys action

.github/workflows/ci_cd.yml
```diff
@@ -77,6 +77,19 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
 
+  check-vulnerabilities:
+    name: "Check library vulnerabilities (development mode)"
+    if: github.ref != 'refs/heads/main'
+    runs-on: ubuntu-latest
+    steps:
```
RobPasMue, line 84: Add also the main mode

PR #722, opened 2024-04-26
fix: Restrict `protobuf` <5

### LINKED ISSUE #684: Bug: Protobuf incompatible issue
PyMechanical does not have upper limit on probuf which creates issue in python3.9 where 5.26.1 is getting installed hence creating issue when installing other dependencies such as `ansys-api-mechanical` and `ansys-api-platform-instancemanagement`

dipinknair:
Closes #684

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ dependencies = [
     "click>=8.1.3", # for CLI interface
     "clr-loader==0.2.6",
     "grpcio>=1.30.0",
-    "protobuf>=3.12.2",
+    "protobuf>=3.12.2,<5",
     "tqdm>=4.45.0",
 ]
 


pyproject.toml
```diff
@@ -38,7 +38,7 @@ dependencies = [
     "click>=8.1.3", # for CLI interface
     "clr-loader==0.2.6",
     "grpcio>=1.30.0",
-    "protobuf>=3.12.2",
+    "protobuf>=3.12.2,<5",
```
RobPasMue, line 41: Interesting... Any reason why? There shouldn't be a limitation anymore on ansys-api-product-instancemanager

dipinknair: Will update this once https://github.com/ansys/ansys-api-mechanical/pull/3 get merged. Because of `ansys-api-mechanical` restricts <5 I can still see the issue with python 3.9

RobPasMue (approved):
LGTM!

PR #729, opened 2024-05-06
feat: Add tests for animation exports

dipinknair:
Add test for animation exports for all formats


tests/embedding/test_image_animation.py
```diff
@@ -0,0 +1,106 @@
+# Copyright (C) 2022 - 2024 ANSYS, Inc. and/or its affiliates.
```
koubaa, line 1: test_graphics_export seems a better name

tests/embedding/test_image_animation.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+@pytest.mark.embedding
+def test_image_animation(printer, selection, embedded_app):
+    """Tests to check image and animation exports
+
+    ref:  Mechanical/QK_ENG_WB2/QK_ENG_WB2_005
+    """
+    globals().update(global_variables(embedded_app, True))
+    Model.AddStaticStructuralAnalysis()
```
koubaa, line 45: geometry import here seems unnecessary. Start with a mechdb where everything up to the graphics export is done

tests/embedding/test_image_animation.py
```diff
@@ -0,0 +1,106 @@
...
+                file.read()
+            except Exception as e:
+                assert False, f"Failed to read file {image_file}: {e}"
+        os.remove(image_file)
+
+    animation_formats = ["GIF", "AVI", "MP4", "WMV"]
+    settings_720p = Ansys.Mechanical.Graphics.AnimationExportSettings()
+    settings_720p.Width = 1280
+    settings_720p.Height = 720
+    for animation_format in animation_formats:
```
koubaa, line 93: might be better to parameterize the test than have a loop here

koubaa: @dipinknair it is ok to open the mechdb multiple times. That is not expensive for a reasonably sized model.

tests/embedding/test_graphics_export.py
```diff
@@ -0,0 +1,99 @@
...
+@pytest.mark.embedding
+@pytest.mark.parametrize("image_formats", ["PNG", "JPG", "BMP"])
+def test_image_export(printer, embedded_app, image_formats):
+    """Tests to check image export."""
+    printer(f"{image_formats} export")
+    globals().update(global_variables(embedded_app, True))
+    mechdb_file = download_file("graphics_test.mechdb", "pymechanical", "test_files")
+    embedded_app.open(mechdb_file)
+    image_settings = Ansys.Mechanical.Graphics.GraphicsImageExportSettings()
+    image_format = getattr(
```
koubaa, line 47: why a `getattr`?

tests/embedding/test_graphics_export.py
```diff
@@ -0,0 +1,99 @@
...
+from ansys.mechanical.core.examples import delete_downloads, download_file
+
+
+@pytest.mark.embedding
+@pytest.mark.parametrize("image_formats", ["PNG", "JPG", "BMP"])
+def test_image_export(printer, embedded_app, image_formats):
+    """Tests to check image export."""
+    printer(f"{image_formats} export")
+    globals().update(global_variables(embedded_app, True))
+    mechdb_file = download_file("graphics_test.mechdb", "pymechanical", "test_files")
```
koubaa, line 44: @dipinknair if we continue this pattern of associating a mechdb with a test can we please add a fixture where we can put the path to the mechdb in the decorator?

tests/embedding/test_graphics_export.py
```diff
@@ -0,0 +1,99 @@
...
+    image_settings.Height = 720
+
+    STAT_STRUC = Model.Analyses[0]
+    STAT_STRUC_SOLN = STAT_STRUC.Solution
+    DEF = STAT_STRUC_SOLN.Children[1]
+    Tree.Activate([DEF])
+    ExtAPI.Graphics.Camera.SetFit()
+    image_file = os.path.join(os.getcwd(), f"image.{image_format}")
+    ExtAPI.Graphics.ExportImage(image_file, image_format, image_settings)
+    with open(image_file, "rb") as file:
```
koubaa, line 62: add a helper function `is_readable(path)`

tests/embedding/test_graphics_export.py
```diff
@@ -0,0 +1,99 @@
...
+    ExtAPI.Graphics.Camera.SetFit()
+    animation_file = os.path.join(os.getcwd(), f"animation.{animation_format}")
+    DEF.ExportAnimation(animation_file, animation_format, animation_settings)
+    with open(animation_file, "rb") as file:
+        try:
+            file.read()
+        except Exception as e:
+            assert False, f"Failed to read file {animation_file}: {e}"
+    os.remove(animation_file)
+    delete_downloads()
```
koubaa, line 99: don't delete downloads at the end of the test - that will force a re-download each time. no problem with having this persist in the user's appdir, right?

tests/embedding/test_graphics_export.py
```diff
@@ -0,0 +1,99 @@
...
+        Ansys.Mechanical.DataModel.Enums.GraphicsAnimationExportFormat, animation_formats
+    )
+    animation_settings.Width = 1280
+    animation_settings.Height = 720
+
+    STAT_STRUC = Model.Analyses[0]
+    STAT_STRUC_SOLN = STAT_STRUC.Solution
+    DEF = STAT_STRUC_SOLN.Children[1]
+    Tree.Activate([DEF])
+    ExtAPI.Graphics.Camera.SetFit()
```
koubaa, line 90: `Tree.Activate([Model.Analyses[0].Solution.Children[1])`

or...

if there's just one total deformation you can get it by type.

tests/embedding/test_graphics_export.py
```diff
@@ -0,0 +1,99 @@
...
+    mechdb_file = download_file("graphics_test.mechdb", "pymechanical", "test_files")
+    embedded_app.open(mechdb_file)
+    animation_settings = Ansys.Mechanical.Graphics.AnimationExportSettings()
+    animation_format = getattr(
+        Ansys.Mechanical.DataModel.Enums.GraphicsAnimationExportFormat, animation_formats
+    )
+    animation_settings.Width = 1280
+    animation_settings.Height = 720
+
+    STAT_STRUC = Model.Analyses[0]
```
koubaa, line 86: pep8 says variables should be `snake_case`

tests/embedding/test_graphics_export.py
```diff
@@ -0,0 +1,99 @@
...
+    from ansys.mechanical.core import global_variables
+except:
+    # No embedding - this import breaks test collection
+    global_variables = {}
+
+from ansys.mechanical.core.examples import delete_downloads, download_file
+
+
+@pytest.mark.embedding
+@pytest.mark.parametrize("image_formats", ["PNG", "JPG", "BMP"])
```
koubaa, line 39: I prefer the singular form `image_format`

tests/embedding/test_graphics_export.py
dipinknair: Yes.

tests/conftest.py
```diff
@@ -233,6 +234,14 @@ class TestEnv:
     # print(f"deleted virtual environment in {venv_dir}\n")
 
 
+@pytest.fixture(scope="session")
```
dipinknair, line 237: @koubaa I have scoped this fixture to session so that mechdb will be reusable.

tests/conftest.py
```diff
@@ -233,6 +234,14 @@ class TestEnv:
...
+def graphics_test_mechdb_file():
+    """Download mechdb files for graphics export test."""
+    mechdb_file = download_file("graphics_test.mechdb", "pymechanical", "test_files")
+    yield mechdb_file
+    os.remove(mechdb_file)
```
koubaa, line 242: @dipinknair why remove it here? Can't it be reused so it doesn't download again each time the test is run?

koubaa (approved):

tests/conftest.py
dipinknair: @koubaa , Oh I thought you suggested only to reuse for Mutiple tests in same session. I will remove this line of code so that it can be reusable for multiple sessions when we run pytest.

PR #735, opened 2024-05-15
add feature flags to ansys-mechanical cli

koubaa:


dipinknair (approved):

src/ansys/mechanical/core/run.py
```diff
@@ -102,6 +103,13 @@ def _run(args, env, check=False, display=False):
     type=int,
     help="Start mechanical in server mode with the given port number",
 )
+@click.option(
+    "--features",
+    type=str,
+    default=None,
```
dipinknair, line 109: https://mechanical.docs.pyansys.com/version/stable/getting_started/running_mechanical.html , docs need update

dipinknair (changes requested):

src/ansys/mechanical/core/run.py
koubaa: @dipinknair thanks. I think we should split out a separate section for running mechanical using ansys-mechanical. The title of that document is "launching Pymechanical" which isn't exactly right.

dipinknair: Yes. I also like the idea of having separate section for `ansys-mechanical`

dipinknair (approved):

src/ansys/mechanical/core/run.py
koubaa: @dipinknair I filed an issue for having a separate doc section

dipinknair (approved):

PR #739, opened 2024-05-20
feat: Add test for deprecation warning

dipinknair:
add additional case for deprecation warning

### DIFF
--- a/tests/embedding/test_app.py
+++ b/tests/embedding/test_app.py
@@ -44,11 +44,9 @@ def test_app_repr(embedded_app):
 @pytest.mark.embedding
 @pytest.mark.minimum_version(241)
 def test_deprecation_warning(embedded_app):
-    struct = embedded_app.Model.AddStaticStructuralAnalysis()
-    with pytest.warns(UserWarning):
-        struct.SystemID
     harmonic_acoustic = embedded_app.Model.AddHarmonicAcousticAnalysis()
     with pytest.warns(UserWarning):
+        harmonic_acoustic.SystemID
         harmonic_acoustic.AnalysisSettings.MultipleRPMs = True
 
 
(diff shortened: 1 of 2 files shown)


tests/embedding/test_app.py
```diff
@@ -47,6 +47,9 @@ def test_deprecation_warning(embedded_app):
     struct = embedded_app.Model.AddStaticStructuralAnalysis()
     with pytest.warns(UserWarning):
         struct.SystemID
+    harmonic_acoustic = embedded_app.Model.AddHarmonicAcousticAnalysis()
+    with pytest.warns(UserWarning):
```
koubaa, line 51: minor - you can get SystemID from a harmonic acoustic.  So you don't really need a static structural anymore

koubaa (approved):

tests/embedding/test_app.py
dipinknair: Done

tests/embedding/test_app.py
```diff
@@ -44,9 +44,10 @@ def test_app_repr(embedded_app):
 @pytest.mark.embedding
 @pytest.mark.minimum_version(241)
 def test_deprecation_warning(embedded_app):
-    struct = embedded_app.Model.AddStaticStructuralAnalysis()
+    harmonic_acoustic = embedded_app.Model.AddHarmonicAcousticAnalysis()
     with pytest.warns(UserWarning):
-        struct.SystemID
+        harmonic_acoustic.SystemID
+        harmonic_acoustic.AnalysisSettings.MultipleRPMs = True
```
koubaa, line 50: minor - putting both in a `pytest.warns` will technically allow the test to pass if one of these does not cause a deprecation warning. A better design would be to put each in its own `pytest.warns` block

dipinknair: I already merged this. But I will make these changes in the other PR

PR #740, opened 2024-05-20
fix: adding mechanical libraries

dipinknair:
fix `add_mechanical_python_libraries` to take the app itself rather than the version number.


src/ansys/mechanical/core/embedding/app_libraries.py
```diff
@@ -51,11 +51,9 @@
 import os
 import sys
 
-from ansys.tools.path.path import _get_unified_install_base_for_version
 
-
-def add_mechanical_python_libraries(version: int):
```
koubaa, line 57: for backward compatibility, make the argument app_or_version and take both types. If a version is supplied, add an obsolete warning

koubaa (approved):

PR #750, opened 2024-05-27
MAINT: Delete the apt-get lists after installing packages

MaxJPRey:
Cleaning up the apt cache and removing /var/lib/apt/lists helps keep the image size down. Since the RUN statement starts with apt-get update, the package cache will always be refreshed prior to apt-get install.

Clean up must be performed in the same RUN step, otherwise it will affect image size.

### DIFF
--- a/doc/changelog.d/750.changed.md
+++ b/doc/changelog.d/750.changed.md
@@ -1,0 +1,1 @@
+MAINT: Delete the apt-get lists after installing packages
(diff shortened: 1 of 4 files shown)


RobPasMue (approved):
LGTM!

RobPasMue:
@dipinknair - @MaxJPRey's change will precisely do that, reduce the image size

PR #753, opened 2024-05-30
feat: raise an exception if port or input script aren't provided in batch mode

### LINKED ISSUE #659: ansys-mechanical script in batch mode should require either `-i` or `-port`
Running mechanical in batch mode without a port or a script doesn't make sense. The CLI to do that should enforce this.

klmcadams:
As the title says. Closes #659

### DIFF
--- a/src/ansys/mechanical/core/run.py
+++ b/src/ansys/mechanical/core/run.py
@@ -111,7 +111,7 @@ def _cli_impl(
             raise Exception("Cannot open in server mode with an input script.")
 
     # If the input_script and port are missing in batch mode, raise an exception
-    if (not graphical) and (input_script is None) and (port is None):
+    if (not graphical) and (input_script is None) and (not port):
         raise Exception("An input script, -i, or port, --port, are required in batch mode.")
 
     args = [exe, "-DSApplet"]
--- a/src/ansys/mechanical/core/run.py
+++ b/src/ansys/mechanical/core/run.py
@@ -110,6 +110,10 @@ def _cli_impl(
         if input_script:
             raise Exception("Cannot open in server mode with an input script.")
 
+    # If the input_script and port are missing in batch mode, raise an exception
+    if (not graphical) and (input_script is None) and (port is None):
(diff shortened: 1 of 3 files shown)


src/ansys/mechanical/core/run.py
```diff
@@ -110,6 +110,10 @@ def _cli_impl(
         if input_script:
             raise Exception("Cannot open in server mode with an input script.")
 
+    # If the input_script and port are missing in batch mode, raise an exception
+    if (not graphical) and (input_script is None) and (port is None):
```
koubaa, line 114: input script `or` port, not `and`

koubaa:
please add unit tests to cover the expectations

src/ansys/mechanical/core/run.py
klmcadams: @koubaa if I change it to `if (not graphical) and (input_script is None) or (port is None):`, then it'll raise the exception even when port or input_script is provided. Is that what you want to happen?

klmcadams: @koubaa I just added unit tests to cover those cases. I also had to add the port argument to other batch cli tests so that the exception wouldn't get raised

src/ansys/mechanical/core/run.py
```diff
@@ -110,6 +110,10 @@ def _cli_impl(
         if input_script:
             raise Exception("Cannot open in server mode with an input script.")
 
+    # If the input_script and port are missing in batch mode, raise an exception
+    if (not graphical) and (input_script is None) and (not port):
+        raise Exception("An input script, -i, or port, --port, are required in batch mode.")
+
```
RobPasMue, line 116: Why not just put them as required arguments via click? Click will take care of this for you... no need to do the if.

See https://click.palletsprojects.com/en/8.1.x/options/#basic-value-options

```py
@click.option(..., required=True,...)
```

RobPasMue (changes requested):

RobPasMue (approved):

koubaa (approved):

PR #757, opened 2024-06-06
feat: use changelog.rst instead of CHANGELOG.md for release notes

klmcadams:
Use `changelog.rst` file to create a "Release notes" tab in the documentation instead of using `CHANGELOG.md` to record release notes


doc/source/changelog.rst
```diff
@@ -0,0 +1,922 @@
+.. _ref_release_notes:
+
+Release notes
+#############
+
+This document contains the release notes for the project.
+
+.. vale off
```
dipinknair, line 8: @klmcadams  should we turn on vale at the end?

dipinknair (approved):
LGTM !

doc/source/changelog.rst
klmcadams: Ohh yeah, it should be right after the towncrier release notes start line. I will add it

klmcadams: You're right, it should be at the end. I was referring to the setup guide oops

PR #758, opened 2024-06-06
Doc: Add embedding api references

dipinknair:
Add Embedding API references in Documentation

image


koubaa (approved):

klmcadams:
@dipinknair I think we should implement sphinx-autoapi instead using `ansys-sphinx-theme[autoapi]==0.16.5`. Using autoapi should generate an API reference for all files in the src directory. I can help implement it if you want

PR #768, opened 2024-06-12
fix lifetime issue

koubaa:
and rework test assets folder


koubaa:
@dipinknair after this lands, we should make a new release on pypi

tests/assets/Eng157.x_t
```diff
@@ -67,4 +67,4 @@ T51 : TRANSMIT FILE created by modeller version 110015017 SCH_1100150_110040 12
  50 93 61 0 53 0 88 0 +0 .0127 0 0 0 -1 -1 0 0 81 1 91 125 89 53 0 0 87 0 104 83
  3 104 .752941176470588 .752941176470588 .752941176470588 83 3 92 .7529411764705
```
koubaa, line 68: @dipinknair @klmcadams maybe this is OK but we probably should prevent the style precommits from changing CAD files like these parasolid ones

tests/assets/cube-hole.mechdb
dipinknair, line 1: Better to add mechdb here in https://github.com/ansys/example-data/tree/master/pymechanical/test_files ?

koubaa: @dipinknair this one is tiny so not really needed in this case but in general yes

Issue #738, opened 2024-05-20
Add option to not copy profiles in appdata.py


klmcadams:
### 📝 Description of the feature

By default, we copy the contents of the appdata folders when we set private_appdata as True. Let's add an option to not run copy_profiles()

### 💡 Steps for implementing the feature

Change the initialize function in the appdata.py file to only run copy_profiles() under specific circumstances

PR #770, opened 2024-06-15
Feat: Add option not to copy app data all the time

AdvaitChandorkar07:
As mentioned in issue https://github.com/ansys/pymechanical/issues/738. Provided an option not to copy profiles by default it will as the code used to


RobPasMue:
Looks like a good idea @ansys/pymechanical-maintainers

dipinknair (approved):
Thanks for the opening this PR @AdvaitChandorkar07  !  LGTM.

PR #772, opened 2024-06-17
fix pythonnet issue

koubaa:
This fixes the nightly runs issue. It's a weird failure mode...

I added the event handlers for updating the globals which doesn't work when running on pythonnet. Since this happens in the constructor, this will throw an exception and the atexit handler will be called, which also throws an exception. Apparently, this will cause the asyncio-based subprocess handling to hang. I'm not sure why this only seems to affect 242...

### DIFF
--- a/doc/changelog.d/772.fixed.md
+++ b/doc/changelog.d/772.fixed.md
@@ -1,0 +1,1 @@
+fix pythonnet issue
--- a/src/ansys/mechanical/core/embedding/app.py
+++ b/src/ansys/mechanical/core/embedding/app.py
@@ -287,9 +287,14 @@ class App:
         return self._version
 
     def _subscribe(self):
-        self._subscribed = True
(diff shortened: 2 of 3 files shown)


dipinknair (approved):

koubaa:
This _was_ ready to merge, but I'm keeping it open until I look further into why this didn't cause any issues in 241. I'd have expected it to.

koubaa:
241: https://github.com/ansys/pymechanical/actions/runs/9551990561/job/26327687609?pr=772
242: https://github.com/ansys/pymechanical/actions/runs/9550352583/job/26322107742

Before the fix, it is not clear why 242 hangs on shutdown but 241 does not. I will test the engrdata serial option

koubaa:
It is not related to the serial option for engineering data, so I don't know the cause. Merging the fix as-is

PR #783, opened 2024-06-21
Fix: Building gallery check

dipinknair:


src/ansys/mechanical/core/embedding/app.py
```diff
@@ -334,6 +332,12 @@ def _update_all_globals(self) -> None:
     def _print_tree(self, node, max_lines, lines_count, indentation):
         """Recursively print till provided maximum lines limit.
 
+        Each object in the tree is expected to have the following attributes:
+         - Name: The name of the object.
+         - Suppressed : Print as suppressed, if object is suppressed.
+         - Children: Checks if object have children.
```
sbradrick, line 338: Children: has children (or maybe "includes children").

src/ansys/mechanical/core/embedding/app.py
```diff
@@ -334,6 +332,12 @@ def _update_all_globals(self) -> None:
...
+           Each child node is expected to have the all these attributes.
```
sbradrick, line 339: Each child node is expected to have all these attributes.

PR #784, opened 2024-06-21
fix sharing app instances, clarify contract

koubaa:
Tested with embedded examples - https://github.com/ansys/pymechanical-embedding-examples/pull/203

### DIFF
--- a/src/ansys/mechanical/core/embedding/app.py
+++ b/src/ansys/mechanical/core/embedding/app.py
@@ -121,13 +121,13 @@ class App:
 
         if BUILDING_GALLERY:
             if len(INSTANCES) != 0:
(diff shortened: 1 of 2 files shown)


dipinknair (approved):

src/ansys/mechanical/core/embedding/app.py
```diff
@@ -297,8 +297,8 @@ def _share(self, other) -> None:
         they all point to the same underlying application
         object. Because of that, special care needs to be
         taken to properly share the state. Other will be
```
sbradrick, line 299: they all point to the same underlying application object. As a result, you need to be careful to properly share the state.

PR #787, opened 2024-06-24
FEAT: Add test for building gallery

dipinknair:
When building gallery only one instance of pymechanical app should be used.


koubaa (changes requested):
do this in embedding_scripts, not embedding. We don't wanto to affect the test fixture behavior for embedding

tests/embedding/test_app.py
```diff
@@ -262,6 +262,32 @@ def test_normal_appdata(pytestconfig, run_subprocess, rootdir):
     assert "ShowTriad value is False" in stdout
 
 
+@pytest.mark.embedding_scripts
+@pytest.mark.python_env
```
koubaa, line 266: remove the python env mark

tests/scripts/build_gallery_test.py
```diff
@@ -0,0 +1,57 @@
...
+import sys
+
+import ansys.mechanical.core as pymechanical
+
+
+def launch_app(version):
+    """Launch embedded instance of app."""
+    # Configuration.configure(level=logging.DEBUG, to_stdout=True, base_directory=None)
+    app = pymechanical.App(version=version)
+    app.update_globals(globals())
```
koubaa, line 33: no need to update globals

tests/scripts/build_gallery_test.py
```diff
@@ -0,0 +1,57 @@
...
+    version = int(sys.argv[1])
+    build_gallery_flag = sys.argv[2]
+    app1 = launch_app(version)
+
+    if build_gallery_flag == "True":
+
+        pymechanical.BUILDING_GALLERY = True
+
+        try:
+            app2 = launch_app(version)
```
koubaa, line 47: app2 should be enough

tests/scripts/build_gallery_test.py
```diff
@@ -0,0 +1,57 @@
...
+            app3 = launch_app(version)
+            print("Multiple App launched with building gallery flag on")
+        except Exception as e:
```
koubaa, line 50: no need for a try/except here, when building gallery we don't expect any exception. The principle here is "Don't expect the unexpected"

tests/embedding/test_app.py
```diff
@@ -262,6 +262,32 @@ def test_normal_appdata(pytestconfig, run_subprocess, rootdir):
...
+    and will fail on PyMechanical version 0.11.0
+    """
+    version = pytestconfig.getoption("ansys_version")
+
+    embedded_gallery_py = os.path.join(rootdir, "tests", "scripts", "build_gallery_test.py")
+
+    stdout, _ = run_subprocess([sys.executable, embedded_gallery_py, version, "False"])
+    stdout = stdout.decode()
+    # Assert Exception
+    assert "Cannot have more than one embedded mechanical instance" in stdout
```
koubaa, line 282: shouldn't this be in the stderr?

koubaa (changes requested):

tests/scripts/build_gallery_test.py
dipinknair: on embedded examples repo. app.update_globals was causing issue when building gallery. 
image

koubaa: oh, right. disregard

koubaa (approved):

PR #797, opened 2024-06-28
FEAT: Integrate ansys visualization tool

dipinknair:
- Replace pyvista plotter with ansys visualization plotter
 - add func to return plotter object.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ build-backend = "flit_core.buildapi"
 [project]
 # Check https://flit.readthedocs.io/en/latest/pyproject_toml.html for all available sections
 name = "ansys-mechanical-core"
-version = "0.11.dev0"
+version = "0.12.dev0"
 description = "A python wrapper for Ansys Mechanical"
 readme = "README.rst"
 requires-python = ">=3.9,<4.0"
@@ -80,7 +80,7 @@ doc = [
     "sphinxemoji==0.3.1",
 ]
 viz = [
-    "pyvista>=0.39.1",
+    "ansys-tools-visualization-interface==0.2.6",
     "usd-core==24.3",
 ]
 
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -80,7 +80,7 @@ doc = [
     "sphinxemoji==0.3.1",
 ]
 viz = [
-    "pyvista>=0.39.1",
+    "ansys-tools-visualization-interface==0.2.6",
```
RobPasMue, line 83: I wouldn't hardcode a specific version - just use above this one

src/ansys/mechanical/core/embedding/viz/pyvista_plotter.py
```diff
@@ -57,7 +58,7 @@ def _get_nodes_and_coords(tri_tessellation: "Ansys.Mechanical.Scenegraph.TriTess
 
 def to_pyvista_plotter(app: "ansys.mechanical.core.embedding.App"):
```
AlejandroFernandezLuces, line 59: This function should be renamed, since we are not using `pyvista`.

src/ansys/mechanical/core/embedding/viz/pyvista_plotter.py
```diff
@@ -68,9 +69,3 @@ def to_pyvista_plotter(app: "ansys.mechanical.core.embedding.App"):
         color = pv.Color(bgr_to_rgb_tuple(body.Color))
         plotter.add_mesh(polydata, color=color, smooth_shading=True)
```
AlejandroFernandezLuces, line 70: You probably want to use `plot` method here, extra kwargs are passed downstream to the PyVista backend.
```suggestion
        plotter.plot(polydata, color=color, smooth_shading=True)
```

AlejandroFernandezLuces (changes requested):
Thanks for the integration of the visualization interface 🙂 

A couple of comments, I see this is still in draft, feel free to ping me again when needed.

PR #802, opened 2024-07-02
feat: add --script-args argument to ansys-mechanical

### LINKED ISSUE #483: Add option to send arbitrary additional arguments with ansys-mechanical
also include the option to add arguments

 -ScriptArgs "arg1, arg2, arg3"

leads to global variables "args" with the value ["arg1", "arg2", "arg3"]

klmcadams:
Allows user to add -ScriptArgs in Mechanical using ansys-mechanical when running a script

Closes #483

### DIFF
--- a/src/ansys/mechanical/core/run.py
+++ b/src/ansys/mechanical/core/run.py
@@ -117,7 +117,9 @@ def _cli_impl(
 
     if script_args:
         if '"' in script_args:
-            raise Exception("Cannot have double quotes in the --script-args string.")
+            raise Exception(
+                "Cannot have double quotes around individual arguments in the --script-args string."
+            )
 
     # If the input_script and port are missing in batch mode, raise an exception
     if (not graphical) and (input_script is None) and (not port):
--- a/src/ansys/mechanical/core/run.py
+++ b/src/ansys/mechanical/core/run.py
@@ -91,6 +91,7 @@ def _cli_impl(
     port: int = 0,
     debug: bool = False,
     input_script: str = None,
+    script_args: str = None,
     exe: str = None,
     version: int = None,
     graphical: bool = False,
@@ -110,6 +111,9 @@ def _cli_impl(
             raise Exception("Cannot open in server mode with a project file.")
         if input_script:
             raise Exception("Cannot open in server mode with an input script.")
+
+    if not input_script and script_args:
+        raise Exception("Cannot add script arguments without an input script.")
 
     # If the input_script and port are missing in batch mode, raise an exception
     if (not graphical) and (input_script is None) and (not port):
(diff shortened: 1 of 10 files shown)


src/ansys/mechanical/core/run.py
```diff
@@ -142,6 +146,13 @@ def _cli_impl(
         args.append("-script")
         args.append(input_script)
 
+    if script_args:
+        args.append("-ScriptArgs")
+        if '"' in script_args:
```
koubaa, line 151: I don't fully understand the purpose of this if expression

src/ansys/mechanical/core/run.py
```diff
@@ -218,6 +229,11 @@ def _cli_impl(
     default=None,
     help="Name of the input Python script. Cannot be mixed with -p",
 )
+@click.option(
+    "--script-args",
+    default=None,
+    help="Arguments to pass into the --input-script, -i. Can only be used with -i.",
```
koubaa, line 235: need to document the format. Are they comma delimited, semicolon delimited, etc

tests/test_cli.py
```diff
@@ -112,6 +112,21 @@ def test_cli_script(disable_cli):
     assert "foo.py" in args
 
 
+@pytest.mark.cli
+def test_cli_scriptargs(disable_cli):
+    args, _ = _cli_impl(
+        exe="AnsysWBU.exe",
+        version=241,
+        input_script="foo.py",
+        script_args="arg1, arg2, arg3",
```
koubaa, line 121: remove the spaces in the string, that's more conventional for this kind of thing

src/ansys/mechanical/core/run.py
koubaa: @klmcadams yes it has to be wrapped in quotes. Is that what this is doing?

klmcadams: @koubaa It checks if there's a double quote in the script_args, and if so, it'll wrap it in single quotes. Otherwise, it'll wrap it in double quotes

If --script-args='"arg1",arg2', then it should wrap the arg in single quotes since the script_args contain double quotes. But for --script_args="arg1,arg2", it'll wrap it in double quotes

koubaa: @klmcadams can you add unit tests to cover these cases?

src/ansys/mechanical/core/run.py
```diff
@@ -111,6 +112,13 @@ def _cli_impl(
         if input_script:
             raise Exception("Cannot open in server mode with an input script.")
 
+    if not input_script and script_args:
+        raise Exception("Cannot add script arguments without an input script.")
+
+    if script_args:
+        if '"' in script_args:
```
koubaa, line 119: @klmcadams if I do this:

`ansys-mechanical -i script.py -script-args "arg1,arg2,arg3"`

then there are double quotes around the script args. Will this work? I think a lot of people are in the habit of putting double quotes around command line arguments that are long

klmcadams: It will work! Whenever it reads the script-args string in, it removes the double quotes so it just becomes `arg1,arg2,arg3` which is why we have to add the double quotes around the args again whenever we append it to the args list. I updated the exception line to specify that you can't add double quotes around individual arguments within the --script-args string to clarify

koubaa (approved):

PR #837, opened 2024-07-29
MAINT: Update ``shims.material_import``

dipinknair:
Update with deprecation warning


src/ansys/mechanical/core/embedding/shims.py
```diff
@@ -1,40 +0,0 @@
...
-but workarounds exist in an older release
-"""
-
-
-def import_materials(
-    app: "ansys.mechanical.core.embedding.Application", material_file: str
-) -> None:
-    """Import material from matml file."""
-    if app._version >= 232:
-        materials = app.DataModel.Project.Model.Materials
```
koubaa, line 35: keep the shims - users might be using them.  just add an obsolete warning to it.

koubaa (changes requested):

src/ansys/mechanical/core/embedding/shims.py
koubaa: This isn't correct, the jscript workaround is what we have to remove!

dipinknair: I have updated and added tests as well.

tests/embedding/test_app.py
```diff
@@ -292,12 +292,19 @@ def test_building_gallery(pytestconfig, run_subprocess, rootdir):
 
 
 @pytest.mark.embedding
-def test_rm_lockfile(embedded_app, tmp_path: pytest.TempPathFactory):
```
koubaa, line 295: why did you remove test_rm_lockfile?

dipinknair: It was not intended, I think I was adding this above it

koubaa (approved):

PR #858, opened 2024-08-07
Refactor usd export

koubaa:
Separate getting the mechanical scene from converting that scene to USD. Because mechanical isn't threadsafe, getting the scene must be done on mechanical's main thread, while converting to USD can be done on any thread.


test.py
dipinknair, line 1: this file might be uploaded by mistake. I think we should have test for both plot and usd_converter

PR #859, opened 2024-08-07
TEST: rpyc 

dipinknair:
Idea is to have remote mechanical embedding services using rpyc


src/ansys/mechanical/core/embedding/rpyc/server.py
```diff
@@ -0,0 +1,96 @@
...
+        self,
+        service: typing.Type[MechanicalService],
+        port: int = 18861,
+        version: int = None,
+        methods: typing.List[typing.Callable] = [],
+        impl=None,
+    ):
+        """Initialize the server."""
+        self._exited = False
+        self._app: mech.App = None
```
koubaa, line 49: use the "BackgroundApp" class to wrap exited, thread, and app.  It'll simplify this class

src/ansys/mechanical/core/embedding/rpyc/server.py
```diff
@@ -0,0 +1,96 @@
...
+from rpyc.utils.server import ThreadedServer
+
+import ansys.mechanical.core as mech
+import ansys.mechanical.core.embedding.utils as utils
+
+from .service import MechanicalService
+
+
+class Server:
+    """Start rpyc server."""
```
koubaa, line 37: rpyc is an implementation detail. Please keep it invisible to the user. We made the same mistake with grpc before

koubaa (changes requested):

PR #882, opened 2024-08-26
feat: launch_gui command

klmcadams:
- Users can call app.launch_gui() to launch their current project in graphical mode
- Uses subprocess.Popen to launch the GUI using the command line given the temporary project file and version of the project


src/ansys/mechanical/core/embedding/app.py
```diff
@@ -191,6 +193,31 @@ def save_as(self, path):
         """Save the project as."""
         self.DataModel.Project.SaveAs(path)
 
+    def launch_gui(self):
+        """Launch the GUI."""
+        # Get the project directory
+        project_directory = self.DataModel.Project.ProjectDirectory
+        # Create a temporary file
+        named_temp_file = tempfile.NamedTemporaryFile()
```
koubaa, line 201: `named_temp_file` is a file that python creates somewhere under `tmp`.  That file isn't used here

src/ansys/mechanical/core/run.py
```diff
@@ -35,7 +35,7 @@
 from ansys.mechanical.core.embedding.appdata import UniqueUserProfile
 from ansys.mechanical.core.feature_flags import get_command_line_arguments, get_feature_flag_names
 
-DRY_RUN = False
+DRY_RUN = False  # control this with an env var to opt into dry run?
```
dipinknair, line 38: delete the comment here ?
```suggestion
DRY_RUN = False

Comment (RobPasMue) [reply src/ansys/mechanical/core/run.py:38]: Or at least mark it as a TODO... ``# TODO: ...``

File: src/ansys/mechanical/core/embedding/ui.py [hunk 3]
```diff
@@ -0,0 +1,185 @@
...
+    """
+    if _is_saved(app):
+        launcher.save_original(app)
+        mechdb_file, temp_file = launcher.save_temp_copy(app)
+        launcher.open_original(app, mechdb_file)
+        p = launcher.graphically_launch_temp(app, mechdb_file, temp_file)
+
+        # If the user wants the temporary file to be deleted
+        if delete_tmp_on_close:
+            atexit.register(launcher._cleanup_gui, p, temp_file)
```
koubaa, line 162: this doesn't need to be onexit, I think _cleanup_gui can be run immediately

tests/embedding/test_app.py
```diff
@@ -323,3 +324,64 @@ def test_app_execute_script(embedded_app):
     with pytest.raises(Exception):
         # This will throw an exception since no module named test available
         embedded_app.execute_script("import test")
+
+
```
koubaa, line 328: add a test for the pid/tempfile cleanup. You don't need to run launch_gui to use that script, you can just start a dummy subprocess that does a 1 second sleep

tests/embedding/test_app.py
```diff
@@ -323,3 +324,64 @@ def test_app_execute_script(embedded_app):
...
+@pytest.mark.embedding
+def test_launch_ui(embedded_app, tmp_path: pytest.TempPathFactory):
+    """Test the _launch_ui function with a mock launcher."""
+
+    class MockLauncher:
```
koubaa, line 333: very good usage/implementation of a mock in a test!

koubaa (changes requested):

src/ansys/mechanical/core/embedding/ui.py
klmcadams: If it's run immediately, wouldn't it affect the launched gui if it's still open and we're deleting the file it's using?

src/ansys/mechanical/core/run.py
klmcadams: Ah, I forgot to delete this comment

tests/embedding/test_app.py
```diff
@@ -323,3 +327,111 @@ def test_app_execute_script(embedded_app):
     with pytest.raises(Exception):
         # This will throw an exception since no module named test available
         embedded_app.execute_script("import test")
+
+
+@pytest.mark.embedding
+def test_launch_ui(embedded_app, tmp_path: pytest.TempPathFactory):
```
klmcadams, line 333: Is this needed since we can test launch_gui in the test_launch_gui test below?

pyproject.toml
```diff
@@ -51,6 +51,7 @@ tests = [
     "pytest==8.3.2",
     "pytest-cov==5.0.0",
     "pytest-print==1.0.1",
+    "psutil==6.0.0"
```
RobPasMue, line 54: Shouldn't it be added to the normal deps as well?

RobPasMue (approved):

tests/embedding/test_app.py
koubaa: @klmcadams it is only needed if we use a "real" embedded app.  We could get away with a fake one here too if we mock out _is_saved.  But I think this is better as-is.

tests/embedding/test_app.py
```diff
@@ -323,3 +327,111 @@ def test_app_execute_script(embedded_app):
...
+    assert m.ops[4] == "open_orig_mechdb"
+    assert m.ops[5] == "launch_temp_mechdb"
+
+
+@pytest.mark.embedding
+def test_launch_gui(embedded_app, tmp_path: pytest.TempPathFactory, capfd):
+    """Test the GUI is launched for an embedded app."""
+    mechdb_path = os.path.join(tmp_path, "test.mechdb")
+    embedded_app.save(mechdb_path)
+    embedded_app.launch_gui(delete_tmp_on_close=False)
```
koubaa, line 398: does this actually start the gui process?

tests/embedding/test_app.py
```diff
@@ -323,3 +327,111 @@ def test_app_execute_script(embedded_app):
...
+
+    # Assert the file and folder exist
+    assert temp_file.exists()
+    assert temp_folder.exists()
+
+    # Run process
+    process, stdout, stderr = run_subprocess(["sleep", "3"])
+
+    # Remove the temporary file and folder
+    sys.argv[1:] = [process.pid, temp_file]
```
koubaa, line 432: don't modify the process wide sys.argv - that could mess up pytest.  Run the `cleanup_gui_main` script as a subprocess here.

tests/embedding/test_app.py
klmcadams: I believe so since `xvfb-run` is being used with pytest

koubaa: I don't think we should do it then, that could be flaky. We should just check that the command line is executed (maybe the dry run mode of run.py) ?

tests/embedding/test_app.py
```diff
@@ -323,3 +327,110 @@ def test_app_execute_script(embedded_app):
...
+
+
+@pytest.mark.embedding
+def test_launch_gui(embedded_app, tmp_path: pytest.TempPathFactory, capfd):
+    """Test the GUI is launched for an embedded app."""
+    mechdb_path = os.path.join(tmp_path, "test.mechdb")
+    embedded_app.save(mechdb_path)
+    embedded_app.launch_gui(delete_tmp_on_close=False, dry_run=True)
+    embedded_app.close()
+    out, err = capfd.readouterr()
```
koubaa, line 400: launch_gui should return the command line if dry_run is true. That way this test can verify it.

koubaa (changes requested):

src/ansys/mechanical/core/embedding/ui.py
```diff
@@ -104,18 +102,22 @@ def graphically_launch_temp(
...
+                    "--graphical",
+                    "--revision",
+                    str(app.version),
+                ],
+            )
 
-        return p
+            return process
+        else:
+            return f"ansys-mechanical --project-file {temp_file} --graphical \
```
koubaa, line 119: minor, you can remove duplication by doing
```
args = [
    "ansys-mechanical",
...
]
if not self._dry_run:
    process = Popen(args)
    return process
else:
    return " ".join(args)
```

koubaa (approved):

PR #902, opened 2024-09-16
FEAT: Add method to execute script from file for embedding

dipinknair:
Leverage `execute_script` method for executing scripts from file

### DIFF
--- a/src/ansys/mechanical/core/embedding/app.py
+++ b/src/ansys/mechanical/core/embedding/app.py
@@ -236,6 +236,13 @@ class App:
             raise Exception(error_msg)
         return script_result.Value
 
+    def execute_script_from_file(self, file_path=None):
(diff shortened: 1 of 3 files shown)


src/ansys/mechanical/core/embedding/app.py
```diff
@@ -236,6 +236,13 @@ def execute_script(self, script: str) -> typing.Any:
             raise Exception(error_msg)
         return script_result.Value
 
+    def execute_script_from_file(self, file_path=None):
+        """Execute the given script from file with the internal IronPython engine."""
+        text_file = open(file_path, "r", encoding="utf-8")
```
koubaa, line 241: any exceptions raised in the file won't have the filename anymore. Maybe that's ok.

koubaa (approved):

PR #935, opened 2024-10-08
feat: `ansys-mechanical-ideconfig` command

klmcadams:
Given an IDE, type of settings to update (user or workspace), and the Mechanical revision number, `ansys-mechanical-ideconfig` will print the location of the settings, along with the content to add to the settings.json files so that autocomplete with `ansys-mechanical-stubs` works


pyproject.toml
```diff
@@ -33,6 +33,7 @@ dependencies = [
     "appdirs>=1.4.0",
     "click>=8.1.3", # for CLI interface
     "clr-loader==0.2.6",
+    "GitPython==3.1.43",
```
koubaa, line 36: shouldn't this be a dependency only for testing?

src/ansys/mechanical/core/autocomplete.py
```diff
@@ -0,0 +1,140 @@
...
+    if settings_type == "user":
+        # Get the path to the user's settings.json file depending on the platform
+        if "win" in sys.platform:
+            settings_json = Path(os.environ.get("APPDATA")) / "Code" / "User" / "settings.json"
+        elif "lin" in sys.platform:
+            settings_json = (
+                Path(os.environ.get("HOME")) / ".config" / "Code" / "User" / "settings.json"
+            )
+    elif settings_type == "workspace":
+        # Find the git repository
```
koubaa, line 68: not every workspace is a git repo

dipinknair:
I think we should capture this in documentation also.

src/ansys/mechanical/core/autocomplete.py
```diff
@@ -0,0 +1,144 @@
...
+import sys
+import sysconfig
+
+import ansys.tools.path as atp
+import click
+
+
+def _cli_impl(
+    ide: str = "vscode",
+    settings_type: str = "user",
```
koubaa, line 37: have a separate function for vscode, so that its more clear that all of the logic is vscode specific right now
```
if ide == ""vscode":
    return _vscode_impl(...args)
else:
    raise Exception(...)
```

pyproject.toml
```diff
@@ -87,6 +87,7 @@ viz = [
 
 [project.scripts]
 ansys-mechanical = "ansys.mechanical.core.run:cli"
+ansys-mechanical-autocomplete = "ansys.mechanical.core.autocomplete:cli"
```
koubaa, line 90: I'm not sure about the entrypoint name. What about
`ansys-mechanical-ideconfig`? (Don't love this either but I think it's more clear)

src/ansys/mechanical/core/autocomplete.py
```diff
@@ -0,0 +1,144 @@
...
+@click.command()
+@click.help_option("--help", "-h")
+@click.option(
+    "--ide",
+    default="vscode",
+    type=str,
+    help="The IDE being used.",
+)
+@click.option(
+    "--settings_type",
```
koubaa, line 104: change this to `--target`

pyproject.toml
klmcadams: Sure, I think that makes more sense too

PR #941, opened 2024-10-10
feat: Add cli command to open documentation

dipinknair:
usage: ``ansys-mechanical --docs``

- [ ] test
- [ ] Should it be different cli like `ansys-mechanical-docs` ?
- [ ] Docs


RobPasMue (commented):
Not a bad idea but... is it necessary to provide a CLI option to open the docs? In any case, my preference is ``ansys-mechanical --docs`` but I don't really know if this will be useful

PR #942, opened 2024-10-11
fix: Warning for multiple version

dipinknair:
use the ``warnings`` module to issue a warning instead of raising it as an exception.

### DIFF
--- a/src/ansys/mechanical/core/embedding/app.py
+++ b/src/ansys/mechanical/core/embedding/app.py
@@ -239,7 +239,7 @@ class App:
         if script_result.Error is not None:
             error_msg += f": {script_result.Error.Message}"
             raise Exception(error_msg)
-        return str(script_result.Value)
+        return script_result.Value
 
     def execute_script_from_file(self, file_path=None):
         """Execute the given script from file with the internal IronPython engine."""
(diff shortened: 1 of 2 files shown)


src/ansys/mechanical/core/embedding/app.py
```diff
@@ -239,7 +239,7 @@ def execute_script(self, script: str) -> typing.Any:
         if script_result.Error is not None:
             error_msg += f": {script_result.Error.Message}"
             raise Exception(error_msg)
-        return script_result.Value
+        return str(script_result.Value)
```
dipinknair, line 242: @koubaa This is modified to be consistent with the remote session for future embedded rpc tests integration.

RobPasMue (approved):

src/ansys/mechanical/core/embedding/app.py
koubaa: @dipinknair I don't agree with this change. The server can do this, the embedding app must not

dipinknair: Yeah, that is also true. Okay we can do that on server then.

dipinknair, line 242:
```suggestion
        return script_result.Value
```

koubaa (approved):

PR #950, opened 2024-10-17
fix: bandit warnings

klmcadams:
- Change `pass` to `raise Exception`
- Add `# no sec` lines to subprocesses where the arguments are known/valid
- Remove `shell=True` from subprocess

### DIFF
--- a/src/ansys/mechanical/core/embedding/ui.py
+++ b/src/ansys/mechanical/core/embedding/ui.py
@@ -22,7 +22,10 @@
 """Run Mechanical UI from Python."""
 
 from pathlib import Path
-from subprocess import Popen
+
+# Subprocess is needed to launch the GUI and clean up the process on close.
+# Excluding bandit check.
+from subprocess import Popen  # nosec
 import sys
 import tempfile
 import typing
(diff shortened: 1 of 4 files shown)


src/ansys/mechanical/core/embedding/ui.py
```diff
@@ -22,7 +22,10 @@
 """Run Mechanical UI from Python."""
 
 from pathlib import Path
-from subprocess import Popen
+
+# Subprocess is needed to launch the GUI and clean up the process on close.
```
koubaa, line 26: @klmcadams why does bandit complain about subprocess??

koubaa (changes requested):

src/ansys/mechanical/core/embedding/ui.py
```diff
@@ -22,7 +22,10 @@
...
+# Excluding bandit check.
+from subprocess import Popen  # nosec
```
RobPasMue, line 28: You need to add the specific bandit code you want to ignore: ``# nosec: B404`` in this case

src/ansys/mechanical/core/embedding/ui.py
RobPasMue: https://bandit.readthedocs.io/en/latest/blacklists/blacklist_imports.html#b404-import-subprocess

B404 - importing subprocess is considered a "risky" condition because of security implications. Damages can be performed on your system - although to be fair you can also do them with pure Python...

Anyway, we should justify the import and usage of subprocess, and make sure that all inputs that are passed to subprocess are controlled by us (avoid user injection).

PR #951, opened 2024-10-17
feat: Add overwrite option for `App.save_as()`

### LINKED ISSUE #900: Add overwrite option to save_as()

Add an option to include `overwrite=True` to `save_as()`. This will allow you to overwrite an existing file when doing `save_as()`

_No response_

_No response_

dipinknair:
deletes existing files so that new files can be saved in that location

Closes #900

### DIFF
--- a/src/ansys/mechanical/core/embedding/app.py
+++ b/src/ansys/mechanical/core/embedding/app.py
@@ -212,23 +212,23 @@ class App:
         if not os.path.exists(path):
             self.DataModel.Project.SaveAs(path)
             return
-        else:
-            if not overwrite:
-                raise Exception(
-                    f"File already exists in {path}, Use ``overwrite`` flag to "
-                    "replace the existing file."
-                )
 
-            file_name = os.path.basename(path)
-            file_dir = os.path.dirname(path)
-            associated_dir = os.path.join(file_dir, os.path.splitext(file_name)[0] + "_Mech_Files")
+        if not overwrite:
+            raise Exception(
+                f"File already exists in {path}, Use ``overwrite`` flag to "
+                "replace the existing file."
+            )
 
-            # Remove existing files and associated folder
-            os.remove(path)
-            if os.path.exists(associated_dir):
-                shutil.rmtree(associated_dir)
-            # Save the new file
-            self.DataModel.Project.SaveAs(path)
+        file_name = os.path.basename(path)
+        file_dir = os.path.dirname(path)
+        associated_dir = os.path.join(file_dir, os.path.splitext(file_name)[0] + "_Mech_Files")
+
+        # Remove existing files and associated folder
+        os.remove(path)
+        if os.path.exists(associated_dir):
+            shutil.rmtree(associated_dir)
+        # Save the new file
+        self.DataModel.Project.SaveAs(path)
 
     def launch_gui(self, delete_tmp_on_close: bool = True, dry_run: bool = False):
         """Launch the GUI."""
--- a/src/ansys/mechanical/core/embedding/app.py
+++ b/src/ansys/mechanical/core/embedding/app.py
@@ -221,7 +221,7 @@ class App:
 
             file_name = os.path.basename(path)
             file_dir = os.path.dirname(path)
-            associated_dir = os.path.join(file_dir, file_name.replace(".mechdb", "_Mech_Files"))
+            associated_dir = os.path.join(file_dir, os.path.splitext(file_name)[0] + "_Mech_Files")
 
             try:
                 # Remove existing files and associated folder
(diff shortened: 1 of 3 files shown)


src/ansys/mechanical/core/embedding/app.py
```diff
@@ -194,9 +196,35 @@ def save(self, path=None):
...
+        the backup file is restored to its original location.
+
+        Parameters
+        ----------
+        path: int, optional
+            The path where file needs to be saved.
+        overwrite: bool, optional
+            Whether the file should be overwritten if it already exists (default is False).
+        """
+        if overwrite:
```
koubaa, line 213: minor: normally when I see code like this:

```
if condition:
    20 lines of code
else:
    1 line of code
return
```

I like to refactor it to:
```
if not condition:
    1 line of code
    return
 
20 lines of code
```

This reduces the indentation and lets you confront the simpler case first, which is more readable in my opinion

src/ansys/mechanical/core/embedding/app.py
```diff
@@ -194,9 +196,35 @@ def save(self, path=None):
...
+            if os.path.exists(path):
```
koubaa, line 214: I think you're missing an else branch here. If the path doesn't exist, it should do the same thing as if overwrite is not set to true.

src/ansys/mechanical/core/embedding/app.py
```diff
@@ -194,9 +196,46 @@ def save(self, path=None):
...
+
+        Parameters
+        ----------
+        path: int, optional
+            The path where file needs to be saved.
+        overwrite: bool, optional
+            Whether the file should be overwritten if it already exists (default is False).
+        """
+        if not os.path.exists(path):
+            self.DataModel.Project.SaveAs(path)
```
koubaa, line 214: return after this so we don't need to have any indent below

src/ansys/mechanical/core/embedding/app.py
```diff
@@ -194,9 +196,46 @@ def save(self, path=None):
...
+                    f"File already exists in {path}, Use ``overwrite`` flag to "
+                    "replace the existing file."
+                )
+
+            temp_dir = tempfile.mkdtemp()
+            temp_file_path = os.path.join(temp_dir, os.path.basename(path))
+
+            try:
+                # Move existing file to temp location
+                shutil.move(path, temp_file_path)
```
koubaa, line 227: @dipinknair why don't we just delete the files in path and save_as to it?

src/ansys/mechanical/core/embedding/app.py
```diff
@@ -194,9 +196,46 @@ def save(self, path=None):
...
+                # Save as new file
+                self.DataModel.Project.SaveAs(path)
```
koubaa, line 229: remember that the outcome of a `save` is both a `mechdb` and an accompanying folder.  We need to overwrite the accompanying folder too

dipinknair: I see. I will update accordingly.

src/ansys/mechanical/core/embedding/app.py
dipinknair: In case if it fails to do a save_as, we should not be losing the existing saved files right ?

koubaa: I'm fine with it. The user said to overwrite, I suppose they must not care too much about losing the files they ask to overwrite

dipinknair: I updated code so that it deletes existing files before saving as new.

src/ansys/mechanical/core/embedding/app.py
koubaa: @dipinknair this is not resolved

src/ansys/mechanical/core/embedding/app.py
```diff
@@ -194,9 +195,43 @@ def save(self, path=None):
...
+
+            try:
+                # Remove existing files and associated folder
+                os.remove(path)
+                if os.path.exists(associated_dir):
+                    shutil.rmtree(associated_dir)
+                # Save the new file
+                self.DataModel.Project.SaveAs(path)
+            except Exception as e:
+                raise e
```
koubaa, line 234: why a try/except just to rethrow the exception? Is this just for debugging?

dipinknair: Was part of previous method of restoring backup. I will update it.

src/ansys/mechanical/core/embedding/app.py
dipinknair: Updated !

koubaa (approved):

Issue #85, opened 2023-01-10
[Embedding] linux crash on shutdown


koubaa:
Embedding on linux (which is not yet pushed to pymechanical) has a crash on shutdown:

The crash occurs in the mono COM interop in `cominterop_rcw_finalizer`  https://github.com/mono/mono/blob/main/mono/metadata/cominterop.c

dipinknair:
Because of this issue, coverage is not captured for some test. The subprocess exits on crash.  The coverage is showing less with Linux as mentioned in #573.

koubaa:
This has been fixed for Mechanical version 2025R1.

PR #965, opened 2024-11-04
CI: Modify how job success is verified for CI/CD 

dipinknair:
Because of PyMechanical crash on exit (Known limitation : #85), the tests are verified using the outputs stored in text file. This will be fixed from next version of Mechanical and hence updating the jobs if version > 24R2.


koubaa:
@dipinknair please also take a look at the `@@success@@` stdout workaround in the tests

koubaa (approved):

PR #974, opened 2024-11-08
FIX: Update embedding script tests 

### LINKED ISSUE #973: Update embedding_script tests
Embedding script tests needs to be updated with new changes where no crash on shutdown in linux. We should re write the tests so that we dont need to check ``@@ success@@`` at the end of the test.

dipinknair:
Fixes #973

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -177,7 +177,7 @@ def mke_app_reset(request):
 _CHECK_PROCESS_RETURN_CODE = os.name == "nt"
 
 # set to true if you want to see all the subprocess stdout/stderr
-_PRINT_SUBPROCESS_OUTPUT_TO_CONSOLE = False
+_PRINT_SUBPROCESS_OUTPUT_TO_CONSOLE = True
 
 
 @pytest.fixture()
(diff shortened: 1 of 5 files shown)


tests/embedding/test_background.py
```diff
@@ -62,8 +62,14 @@ def _run_background_app_test(
     Returns the stderr of the subprocess as a string.
     """
     subprocess_pass_expected = pass_expected
```
koubaa, line 64: how would this look if it was a fixture?

koubaa (approved):

tests/conftest.py
```diff
@@ -177,7 +177,7 @@ def mke_app_reset(request):
 _CHECK_PROCESS_RETURN_CODE = os.name == "nt"
 
 # set to true if you want to see all the subprocess stdout/stderr
-_PRINT_SUBPROCESS_OUTPUT_TO_CONSOLE = False
+_PRINT_SUBPROCESS_OUTPUT_TO_CONSOLE = True
```
dipinknair, line 180:
```suggestion
_PRINT_SUBPROCESS_OUTPUT_TO_CONSOLE = False
```

PR #1007, opened 2024-12-18
feat: Option to ignore lock file on open

### LINKED ISSUE #327: Add an option to ignore the lock file on open
This can be useful for debugging and interactive scripts.  Of course we should warn the user that the project file may get corrupted if they use this option and the lock file is removed manually before opening the project

klmcadams:
- Remove the lock file if it exists before opening the project. By default, `remove_lock=False`
- Emit a warning to the user that the project file can be corrupted if `remove_lock=True`

Closes #327

### DIFF
--- a/src/ansys/mechanical/core/embedding/app.py
+++ b/src/ansys/mechanical/core/embedding/app.py
@@ -224,7 +224,7 @@ class App:
             lock_file = Path(self.DataModel.Project.ProjectDirectory) / ".mech_lock"
             if lock_file.exists():
                 warnings.warn(
-                    "Removing the lock file before opening the project. \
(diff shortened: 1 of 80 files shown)


src/ansys/mechanical/core/embedding/app.py
```diff
@@ -217,8 +218,20 @@ def _dispose(self):
         self._app.Dispose()
         self._disposed = True
 
-    def open(self, db_file):
+    def open(self, db_file, remove_lock=False):
         """Open the db file."""
```
dipinknair, line 222: suggestions: Add doc strings for the parameters for api reference

koubaa (approved):

PR #1017, opened 2025-01-03
feat: add `globals` kwarg to app and adjust `ideconfig` behavior

klmcadams:
- Add `globals` kwarg to App class
- Import globals if the app is initialized

ideconfig
- Raise an exception if the revision number is less than the minimum version in ansys-mechanical-stubs
- Raise a warning if the revision number is greater than the maximum version in ansys-mechanical-stubs. Use the max version to print the autocomplete settings


src/ansys/mechanical/core/embedding/app.py
```diff
@@ -186,13 +196,19 @@ def __init__(self, db_file=None, private_appdata=False, **kwargs):
         runtime.initialize(self._version)
         connect_warnings(self)
         self._poster = None
+        # Set the initialized flag to True
+        self._app._initialized = True
 
         self._disposed = False
         atexit.register(_dispose_embedded_app, INSTANCES)
         INSTANCES.append(self)
```
koubaa, line 204: @klmcadams No need to add a flag. You can check `len(INSTANCES)`

src/ansys/mechanical/core/embedding/app.py
```diff
@@ -218,6 +234,10 @@ def _dispose(self):
         self._app.Dispose()
         self._disposed = True
 
+    def is_initialized(self):
```
koubaa, line 237: this should be a free function. otherwise you need an instance of the app to check if the app is initialized

src/ansys/mechanical/core/embedding/global_importer.py
```diff
@@ -0,0 +1,45 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""Import Mechanical globals."""
+
+import clr
```
koubaa, line 24: missing the exception here if the app hasn't been initialized yet

koubaa:
Need to update examples. one should use the new globals() argument. Another should avoid globals entirely and import the types manually after constructing the app instance

.github/workflows/ci_cd.yml
```diff
@@ -697,7 +697,7 @@ jobs:
 
   coverage:
     name: Merging coverage
-    needs: [remote-connect, embedding-tests, embedding-scripts-tests, launch-tests]
+    needs: [remote-connect, embedding-tests, embedding-scripts-tests, embedding-rpc-tests, launch-tests]
```
koubaa, line 700: nice catch

src/ansys/mechanical/core/embedding/app.py
```diff
@@ -89,6 +89,11 @@ def _start_application(configuration: AddinConfiguration, version, db_file) -> "
         return Ansys.Mechanical.Embedding.Application(db_file)
 
 
+def is_initialized():
+    """Check if the app has been initialized."""
+    return True if len(INSTANCES) != 0 else False
```
koubaa, line 94: `return len(INSTANCES) != 0`

src/ansys/mechanical/core/embedding/app.py
```diff
@@ -192,6 +205,10 @@ def __init__(self, db_file=None, private_appdata=False, **kwargs):
         self._updated_scopes: typing.List[typing.Dict[str, typing.Any]] = []
         self._subscribe()
 
+        globals = kwargs.get("globals")
+        if globals is not None and is_initialized():
```
koubaa, line 209: I don't think the `is_initialized` check is necessary here.  `initializer.initialize` already ran here (see line 187)

koubaa (changes requested):

koubaa (approved):

PR #1055, opened 2025-01-20
FEAT: Rpyc integration

dipinknair:
Implement remote calls for embedded app


dipinknair (commented):
@koubaa Using Background App.

src/ansys/mechanical/core/embedding/background.py
```diff
@@ -82,6 +82,12 @@ def post(self, callable: typing.Callable):
             raise RuntimeError("Cannot use BackgroundApp after stopping it.")
         return BackgroundApp.__poster.post(callable)
 
+    def try_post(self, callable: typing.Callable):
```
koubaa, line 85: minor - you can factor the check into a method used by both `post` and `try_post`

tests/conftest.py
```diff
@@ -288,27 +289,65 @@ def connect_to_mechanical_instance(port=None, clear_on_connect=False):
...
+    from ansys.mechanical.core.embedding.rpc.client import Client
+
+    client = Client("localhost", port)
+    return client
+
+
+def stop_embeddedd_server():
+    global embedded_server
+    if embedded_server is not None:
+        embedded_server.terminate()  # Sends a SIGTERM signal
```
koubaa, line 312: add a TODO at least to have this not use a terminate.

tests/test_mechanical.py
```diff
@@ -37,25 +36,31 @@
...
     assert result == "5"
 
 
 @pytest.mark.remote_session_connect
 def test_run_python_script_success_return_empty(mechanical):
-    result = mechanical.run_python_script("ExtAPI.DataModel.Project")
-    if misc.is_windows():
+    result = str(mechanical.run_python_script("ExtAPI.DataModel.Project"))
+    # TODO: Investigate why the result is different for grpc
+    if misc.is_windows() and not hasattr(mechanical, "_rpc_type"):
```
koubaa, line 47: why hasattr?

dipinknair: Adding attr "_rpc_type"  in fixture of remote mechanical case is not sufficient for all cases. In some test remote_session_connect tests where grpc method is used,  they dont use the fixture and directly create new remote instance. In that case the attr is not available for making special case for rpyc. It was choice either to add this attr in Mechanical class itself or add only for case of rpyc scenario.  we can discuss more about this and probably change current implementation.

koubaa: @dipinknair in that case, put a property on the mechanical class itself, not in the fixture.

.github/workflows/ci_cd.yml
```diff
@@ -283,6 +283,81 @@ jobs:
...
+          apt install lsb-release xvfb software-properties-common -y
+          add-apt-repository ppa:deadsnakes/ppa -y
+          apt install -y python${{ matrix.python-version }} python${{ matrix.python-version }}-venv
+          python${{ matrix.python-version }} -m venv /env
+
+      - name: Install packages for testing
+        run: |
+          . /env/bin/activate
+          pip install --upgrade pip
+          pip install -e .[tests]
```
koubaa, line 314: you can do `.[tests,rpyc]` to install in one line

koubaa (approved):

.github/workflows/ci_cd.yml
```diff
@@ -283,6 +283,81 @@ jobs:
...
+          pip install -e .[rpc]
```
dipinknair, line 315:
```suggestion
          pip install -e .[tests, rpc]
```

.github/workflows/ci_cd.yml
```diff
@@ -283,6 +283,80 @@ jobs:
...
+          apt install lsb-release xvfb software-properties-common -y
+          add-apt-repository ppa:deadsnakes/ppa -y
+          apt install -y python${{ matrix.python-version }} python${{ matrix.python-version }}-venv
+          python${{ matrix.python-version }} -m venv /env
+
+      - name: Install packages for testing
+        run: |
+          . /env/bin/activate
+          pip install --upgrade pip
+          pip install -e .[tests, rpc]
```
dipinknair, line 314:
```suggestion
          pip install -e .[tests,rpc]
```

PR #1060, opened 2025-01-23
feat: Create option for PyPIM to be installed separately

### LINKED ISSUE #419: launch_mechanical has a version argument that is only supported if pypim is configured
it should be usable without pypim.

Furthermore, pim should only be installed using a [pim] option.

`pip install ansys-mechanical-core` => pim not installed

`pip install ansys-mechanical-core[pim]` => pim installed

klmcadams:
- Do not install PyPIM by default for launch_mechanical
- User has to `pip install ansys-mechanical-core[pim]` if they would like to use pim with launch_mechanical

Closes #419


koubaa (approved):

src/ansys/mechanical/core/mechanical.py
```diff
@@ -1951,6 +1964,13 @@ def launch_remote_mechanical(version=None) -> (grpc.Channel, Instance):  # pragm
     -------
         Tuple containing channel, remote_instance.
     """
+    # Display warning if PyPIM is not installed
+    if not HAS_ANSYS_PIM:
+        warnings.warn(
+            "Installation of pim option required! Use pip install ansys-mechanical-core[pim]"
```
RobPasMue, line 1970:
```suggestion
            "Installation of pim option required! Use ``pip install ansys-mechanical-core[pim]``"
```

src/ansys/mechanical/core/pool.py
```diff
@@ -26,7 +26,15 @@
...
-import ansys.platform.instancemanagement as pypim
+try:
+    import ansys.platform.instancemanagement as pypim  # pragma: nocover noqa: F401
+
+    HAS_ANSYS_PIM = True
+    """Whether or not PyPIM exists."""
+except ImportError:
+    HAS_ANSYS_PIM = False
+
+
```
RobPasMue, line 37: Why not have this logic at ``src/ansys/mechanical/__init__.py`` so that you don't duplicate it? And then import ``HAS_ANSYS_PIM`` in the module you need afterwards.

RobPasMue (approved):
LGTM - just a few enhancements

src/ansys/mechanical/core/pool.py
klmcadams: That's a good idea, thanks!

RobPasMue (approved):
LGTM!

PR #1068, opened 2025-01-27
FEAT: Add a Message Manager for App

### LINKED ISSUE #440: Add an API to list messages
messages can be printed like this:
```
>>> _=[print(msg.DisplayString) for msg in app.ExtAPI.Application.Messages]
The shared license is not available to perform the Geometry Attach operation
```

ideally this should be an API on `App`, like app.PrintMessages()

dipinknair:
Fixes #440 

```python
app =App()

# Add a custom message
app.messages.add("info", "User clicked the start button.")

# check number of messages
len(app.messages)
... 1

# returns panda data frame representation
app.messages
... Severity             TimeStamp                   DisplayString Source StringID                                Location RelatedObjects
... 0     INFO  1/30/2025 1:46:14 PM  User clicked the start button.   None     None  Ansys.ACT.Core.Utilities.SelectionInfo             []
# print minimalistic view (severity and message)
app.messages.show()
... Severity: info
... DisplayString: User clicked the start button.

# prints all information including location, time, source etc of all message available
app.messages.show(filter='*")
... TimeStamp: 1/30/2025 1:42:01 PM
... Severity: info
... DisplayString: User clicked the start button.
... Source: None
... StringID: None
... Location: Ansys.ACT.Core.Utilities.SelectionInfo
... RelatedObjects: System.Collections.Generic.List`1[Ansys.Mechanical.DataModel.Interfaces.IDataModelObject]

# gets specific message in datafram row representation
print(app.messages[1].Severity)
... Info

# Remove specific message
app.messages.remove(1)

# Clear all messages
app.messages.clear()
````

### CODE REVIEW

**Comment by koubaa** [comment]
`messages.count()` isn't pythonic. It should be `len(messages)`

**Comment by koubaa** [comment]
An on/off for complete_info seems inflexible. How about this:

`app.messages.show(filter="*") # filter defaults to "severity;message". Accepts a semicolon delimited list of columns or a wildcard.`

File: src/ansys/mechanical/core/embedding/messages.py [hunk 1]
```diff
@@ -0,0 +1,220 @@
...
+            raise ValueError(f"Invalid severity: {severity}")
+
+        _msg = self.Message(text, severity_map[severity.lower()])
+        self._app.ExtAPI.Application.Messages.Add(_msg)
+
+        self._update_messages_cache()
+
+    def show(self, complete_info=False):
+        """Print all messages with full details.
+
```
koubaa, line 126: shouldn't this update the dataframe first?

dipinknair: yes

dipinknair:
> An on/off for complete_info seems inflexible. How about this:
> 
> `app.messages.show(filter="*") # filter defaults to "severity;message". Accepts a semicolon delimited list of columns or a wildcard.`

Added this feature and updated the PR descriptions as well

src/ansys/mechanical/core/embedding/app.py
```diff
@@ -34,6 +34,7 @@
 from ansys.mechanical.core.embedding.addins import AddinConfiguration
 from ansys.mechanical.core.embedding.appdata import UniqueUserProfile
 from ansys.mechanical.core.embedding.imports import global_entry_points, global_variables
+from ansys.mechanical.core.embedding.messages import MessageManager
```
koubaa, line 37: can you import this lazily? I think pandas does something on initializatino which might slow things down

src/ansys/mechanical/core/embedding/app.py
```diff
@@ -437,6 +439,14 @@ def project_directory(self):
         """Returns the current project directory."""
         return self.DataModel.Project.ProjectDirectory
 
+    @property
+    def messages(self):
```
koubaa, line 443: I don't like that `_update_messages_cache()` has to be done explicitly. The message manager should be responsible for keeping itself up to date. Use an event handler

src/ansys/mechanical/core/embedding/messages.py
```diff
@@ -0,0 +1,251 @@
...
+            msg = _Message(row)
+            msg.show(filter)
+
+    def clear(self):
+        """Clear all messages."""
+        self._app.ExtAPI.Application.Messages.Clear()
+        self._update_messages_cache()
+
+
+class _Message:
```
koubaa, line 161: why is this necessary? There is already a message object in the API

koubaa (changes requested):

src/ansys/mechanical/core/embedding/messages.py
dipinknair: We already have all messages stored as dataframe. Isnt it better to use that rather than doing API call for each message?

src/ansys/mechanical/core/embedding/messages.py
```diff
@@ -0,0 +1,190 @@
...
+    def __init__(self, app):
+        """Initialize the message manager."""
+        self._app = app
+
+        # Import necessary classes
+        from Ansys.Mechanical.Application import Message
+        from Ansys.Mechanical.DataModel.Enums import MessageSeverityType
+
+        self._message_severity = MessageSeverityType
+        self._message = Message
```
koubaa, line 50: self._message doesn't seem to be used

dipinknair: self._message its used when adding new message

koubaa (approved):

dipinknair:
@koubaa  When app starts there is already a warning message saying license checkout is delayed. This was causing all the tests to fail because how test is written assuming zero message at the beginning.

PR #1073, opened 2025-01-29
fix: Add sleep to poster test

klmcadams:
Added `utils.sleep(400)` to the `change_name_async` function

The test was hanging at this line in `poster.py`: 
https://github.com/ansys/pymechanical/blob/0f4f184b4c2a63c7d3369a03f053388e80728acf/src/ansys/mechanical/core/embedding/poster.py#L84

I added the `utils.sleep()` function due to the docstring in `poster.post()`:

```
The main thread needs to be receiving posted messages in order for this to work from a background thread. 
Use the `sleep` routine from the `utils` module to make Mechanical available to receive messages.
```


tests/embedding/test_app.py
```diff
@@ -206,6 +205,7 @@ def change_name():
         def raise_ex():
             raise Exception("TestException")
 
+        utils.sleep(400)
```
koubaa, line 208: this isn't correct. We aren't supposed to use utils.sleep in a background thread

PR #1074, opened 2025-01-29
FIX: Add option on rpc server for accepting multiple services

dipinknair:
This way user can call  their own servicemethods with DefaultServiceMethods for a custom server
```
server = MechanicalEmbeddedServer(
        port=_port,
        version=_version,
        impl=[DefaultServiceMethods, ServiceMethods],
    )
```
or 
```
server = MechanicalEmbeddedServer(
        port=_port,
        version=_version,
        impl=ServiceMethods,
    )
```


src/ansys/mechanical/core/embedding/rpc/server.py
```diff
@@ -42,24 +42,28 @@
...
-        self.EMBEDDED = True
+        self._install_classes(impl)
 
     def _install_functions(self, methods):
         """Install the given list of methods."""
+        if not methods:
+            return
         [self._install_function(method) for method in methods]
 
+    def _install_classes(self, impl):
```
koubaa, line 58: imo - class can be either a single class or a list.

koubaa (approved):

PR #1088, opened 2025-02-07
FIX: Embedding rpc server shutdown

### LINKED ISSUE #1077: Crash on shutting down rpc server
Embedded server crashing on exit for pytest. This is due to improper shutdown of rpc server

dipinknair:
Fixes #1077

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -27,6 +27,7 @@ import platform
 import shutil
 import subprocess
 import sys
+import time
 
 import ansys.tools.path as atp
 import grpc
@@ -291,6 +292,7 @@ def connect_to_mechanical_instance(port=None, clear_on_connect=False):
 
 def launch_rpc_embedded_server(port: int, version: int, server_script: str):
     """Start the server as a subprocess using `port`."""
+    global embedded_server
     env_copy = os.environ.copy()
     embedded_server = subprocess.Popen(
         [sys.executable, server_script, str(port), str(version)], env=env_copy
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -338,9 +328,8 @@ def mechanical(pytestconfig, rootdir):
     print(mechanical)
     yield mechanical
     if is_embedded_server:
-        print("stopping embedded server")
+        print("\nStopping embedded server fixture")
         mechanical.exit()
```
koubaa, line 332: you should put some code here to unsure that the subprocess exits. Maybe wait up to 10 seconds and if it isn't exited you can terminate it.  Otherwise the python process could leak if there was a problem with shutting down.

If that happens, I don't know what happens if an exception is thrown in a test fixture after the session scope is done. It should fail the overall result, right? It's worth trying.

koubaa (approved):

tests/conftest.py
```diff
@@ -338,9 +330,19 @@ def mechanical(pytestconfig, rootdir):
...
+        print("\n Stopping embedded server")
+        global embedded_server
         mechanical.exit()
-        stop_embeddedd_server()
+        start_time = time.time()
+        while embedded_server.poll() is None:
+            if time.time() - start_time > 10:
+                try:
+                    embedded_server.terminate()
+                    embedded_server.wait(timeout=2)
```
dipinknair, line 341:
```suggestion
                    embedded_server.wait()
```

tests/conftest.py
dipinknair: updated

PR #1090, opened 2025-02-10
FIX: Remove unnecessary prints from embedding rpc

dipinknair:
- [x] Improves how clients waiting till server is ready
- [x] removes some unwanted prints

### DIFF
--- a/src/ansys/mechanical/core/embedding/rpc/client.py
+++ b/src/ansys/mechanical/core/embedding/rpc/client.py
@@ -27,13 +27,14 @@ import time
 
 import rpyc
 
+from ansys.mechanical.core.embedding.rpc.server import PYMECHANICAL_DEFAULT_RPC_PORT
 from ansys.mechanical.core.mechanical import DEFAULT_CHUNK_SIZE
 
 
 class Client:
     """Client for connecting to Mechanical services."""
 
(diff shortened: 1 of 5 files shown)


src/ansys/mechanical/core/embedding/rpc/server.py
```diff
@@ -67,15 +66,6 @@ def _install_class(self, impl):
             elif methodtype == MethodType.PROP:
                 self._install_property(method, methodname)
 
-    def on_connect(self, conn):
```
dipinknair, line 70: @koubaa  this is not necessary since we are printing the app when we start the server. This is what was causing multiple repeating prints

src/ansys/mechanical/core/embedding/rpc/client.py
```diff
@@ -77,22 +77,19 @@ def __getattr__(self, attr):
...
-                conn.ping()  # Simple ping to check if the connection is healthy
-                conn.close()
-                print("Server is ready to connect")
+                self.connection = rpyc.connect(self.host, self.port)
+                self.connection.ping()
+                print("Server is ready.")
                 break
             except:
-                time.sleep(2)
+                time.sleep(0.1)
```
RobPasMue, line 92: Can we do some kind of exponential back off?

RobPasMue (approved):

PR #1101, opened 2025-02-24
FEAT: Add rpyc option in `launch_mechanical`

dipinknair:
`launch_mechanical` now provides two rpc options such as `grpc` ( `backend = 'mechanical'`) and `rpyc`( `backend = 'python'`) . By default, it is `grpc`
It can be used as given below.
```python
import ansys.mechanical.core as pymechanical

mechanical = pymechanical.launch_mechanical(backend="python")
print(mechanical)
```
with `rpyc` option mechanical exists when python code completes.
with cleanup_on_exit option disabled , user can still use another client to connect the mechanical which is alive.

```python
import ansys.mechanical.core as pymechanical

mechanical = pymechanical.launch_mechanical(backend="python", cleanup_on_exit=False)
```

### DIFF
--- a/doc/changelog.d/1096.documentation.md
+++ b/doc/changelog.d/1096.documentation.md
@@ -1,0 +1,1 @@
+Add "what's new" fragment for the globals parameter in App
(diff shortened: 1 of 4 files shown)


koubaa:
@dipinknair I prefer the option to be like this:

`launch_mechanical(backend=python)`
`launch_mechanical(backend=mechanical)`

That way `rpyc` and `grpc` are implementation details.

koubaa (approved):

PR #1106, opened 2025-02-28
CHORE: Add solution status assert in example

dipinknair:
This also fixes the example which is broken

### DIFF
--- a/examples/embedding_n_remote/embedding_remote.py
+++ b/examples/embedding_n_remote/embedding_remote.py
@@ -260,7 +260,7 @@ pressure.Magnitude.Output.DiscreteValues = output_quantities_2
 Model.Solve(True)
 solution = analysis.Solution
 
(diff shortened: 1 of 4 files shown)


examples/embedding_n_remote/embedding_remote.py
```diff
@@ -257,14 +257,15 @@ def write_file_contents_to_console(path, number_lines=-1):
 # Solve model
 # ~~~~~~~~~~~
 
-Model.Solve()
+Model.Solve(True)
+solution = analysis.Solution
 
+assert str(solution.Status) == "Done"
```
koubaa, line 263: Why convert to string? We should have an enum state.

dipinknair, line 263:
```suggestion
assert solution.Status == SolutionStatusType.Done
```

klmcadams (approved):

PR #1114, opened 2025-03-04
FIX: Example and minor code issues

dipinknair:


src/ansys/mechanical/core/ide_config.py
```diff
@@ -93,6 +93,7 @@ def _vscode_impl(
         If unspecified, it finds the default Mechanical version from ansys-tools-path.
     """
     # Update the user or workspace settings
+    settings_json = None
```
klmcadams, line 96:
```suggestion
    settings_json = "the settings.json file"
```
Set settings_json to "the settings.json file", so if for some reason the target isn't user or workspace, then the message that is printed out says "Update the settings.json file with the following information:"

PR #1154, opened 2025-05-02
fix: Install `libjbig-dev` for quarto build to succeed in 252 documentation build

klmcadams:
apt install libjbig-dev and add the path of the libjbig*.so* files to the LD_LIBRARY_PATH so the quarto cheatsheet build succeeds


klmcadams:
Successful build for 252: https://github.com/ansys/pymechanical/actions/runs/14800467679/job/41557835488

PR #1155, opened 2025-05-02
fix: Update globals when BUILDING_GALLERY is True 

klmcadams:
Update globals when the globals kwarg is provided and BUILDING_GALLERY is True

### DIFF
--- a/tests/scripts/run_embedded_app.py
+++ b/tests/scripts/run_embedded_app.py
@@ -40,14 +40,33 @@ def launch_app(args):
         init_msg = "The app is initialized" if is_initialized() else "The app is not initialized"
         print(init_msg)
 
-    app = pymechanical.App(
-        version=int(args.version),
-        private_appdata=args.private_appdata,
-        copy_profile=False,
-        globals=globals(),
-    )
+    if args.update_globals_kwarg:
+        app = pymechanical.App(
+            version=int(args.version),
+            private_appdata=args.private_appdata,
+            copy_profile=False,
+            globals=globals(),
+        )
+    else:
+        app = pymechanical.App(
+            version=int(args.version),
+            private_appdata=args.private_appdata,
+            copy_profile=False,
+        )
+        app.update_globals(globals())
 
     return app
+
+
+def test_globals(args):
+    """Test ViewOrientationType doesn't throw an error when globals are updated."""
+    if args.build_gallery_flag:
+        pymechanical.BUILDING_GALLERY = True
+        print("Building gallery flag is set to True")
+
+    app = launch_app(args)
+    app.Graphics.Camera.SetSpecificViewOrientation(ViewOrientationType.Iso)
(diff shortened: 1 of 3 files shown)


tests/scripts/run_embedded_app.py
```diff
@@ -40,16 +40,35 @@ def launch_app(args):
...
+            copy_profile=False,
+        )
+        app.update_globals(globals())
 
     return app
 
 
+def test_globals(args):
+    """Test ViewOrientationType doesn't throw an error when globals are updated."""
+    if args.build_gallery_flag:
```
koubaa, line 63: the only test coverage we were missing was when BUILDING_GALLERY was set to True and the second app instance had globals updated.

tests/scripts/run_embedded_app.py
```diff
@@ -40,31 +40,32 @@ def launch_app(args):
...
 def test_globals(args):
     """Test ViewOrientationType doesn't throw an error when globals are updated."""
-    if args.build_gallery_flag:
-        pymechanical.BUILDING_GALLERY = True
-        print("Building gallery flag is set to True")
-
+    # Set BUILDING_GALLERY to True
+    pymechanical.BUILDING_GALLERY = True
+    # Launch an embedded instance of the app without updating globals
+    app = pymechanical.App()
```
koubaa, line 65: needs the version number here. (the other arguments don't matter)

klmcadams, line 65:
```suggestion
    app = pymechanical.App(version=int(args.version))
```

tests/scripts/run_embedded_app.py
```diff
@@ -40,16 +40,36 @@ def launch_app(args):
...
 
 
+def test_globals(args):
+    """Test ViewOrientationType doesn't throw an error when globals are updated."""
+    # Set BUILDING_GALLERY to True
+    pymechanical.BUILDING_GALLERY = True
+    # Launch an embedded instance of the app without updating globals
+    app = pymechanical.App(version=int(args.version))
+    # Launch an embedded instance of the app with updated globals
+    app = launch_app(args)
```
koubaa, line 67: it would be more readable here to do this:
```
app = pymechanical.App(version=int(args.version))
app = pymechanical.App(globals=globals())
```

rather that to reuse launch_app, since the other arguments are not related to this tet.

koubaa (approved):

tests/scripts/run_embedded_app.py
```diff
@@ -74,6 +94,9 @@ def print_showtriad(args):
     parser.add_argument(
         "--test_not_initialized", action="store_true"
     )  # 'store_true' implies default=False
+    parser.add_argument(
```
koubaa, line 97: unused now

koubaa (changes requested):
Remove the unused code in run_embedded_app.py

koubaa (approved):

PR #1156, opened 2025-05-05
test: add test case for analytics capture 

koubaa:
Now that 25.2 nightly builds have stabilized, I can add the test case


MaxJPRey (approved):
LGTM. Thanks @koubaa for your time on this test.

PR #1171, opened 2025-05-21
fix: Add `self._messages` to BUILDING_GALLERY section in app initialization

klmcadams:
Add `self._messages = None` to the BUILDING_GALLERY section of the App initialization so the following AttributeError does not occur when running `make -C doc html`:

```
app.messages.show()

if self.messages is not None:
^^^
AttributeError: 'App' object has no attribute '_messages'. Did you mean: 'messages'?
```

### DIFF
--- a/src/ansys/mechanical/core/embedding/app.py
+++ b/src/ansys/mechanical/core/embedding/app.py
@@ -189,6 +189,9 @@ class App:
         # Get the globals dictionary from kwargs
         globals = kwargs.get("globals")
 
+        # Set messages to None before BUILDING_GALLERY check
+        self._messages = None
+
         # If the building gallery flag is set, we need to share the instance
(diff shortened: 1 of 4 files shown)


src/ansys/mechanical/core/embedding/app.py
```diff
@@ -201,6 +201,8 @@ def __init__(self, db_file=None, private_appdata=False, **kwargs):
                 if globals:
                     # The next line is covered by test_globals_kwarg_building_gallery
                     instance.update_globals(globals)  # pragma: nocover
+                # Set messages to None
+                self._messages = None
```
koubaa, line 205: I prefer to put `self._messages = None` one time before the check for if `BUILDING_GALLERY`

koubaa (changes requested):

koubaa (approved):

PR #1174, opened 2025-05-22
Fix : Embedding message get test

dipinknair:
The failing test uses one of the mechdb to get message from Mechanical which is not consistent with the new version of Mechanical. It is replaced with another mechdb that we use in tests.


dipinknair:
https://github.com/ansys/pymechanical/actions/runs/15187653847
passing with 252.

koubaa (approved):

PR #1268, opened 2025-08-06
FEAT: Add plot option for mesh

koubaa:
Use new mechanical API to get the scenegraph for meshes and results

Add handling for line bodies, point clouds, arbitrary scenegraph structure, and MeshOrientedTransformNode (under an env var, it isn't correct)

To render a mesh:
`app.plot(app.Model.Mesh)`

To render a result:

`app.plot(app.Model.GetChildren(DataModelObjectCategory.Result, True)[0])`


dipinknair:
@koubaa let me test this on examples.

PR #1271, opened 2025-08-07
FIX: Update release action

dipinknair:
This change will not publish pdf or html files to github release. Currently PyMech doesnt build pdf which is blocking this action from releasing.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -808,6 +808,7 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           library-name: ${{ env.PACKAGE_NAME }}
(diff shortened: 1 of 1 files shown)


doc/source/conf.py
```diff
@@ -331,8 +331,11 @@
     "https://support.ansys.com/*",
     "https://discuss.ansys.com/*",
     "https://www.ansys.com/*",
+    "../api/*",  # Remove this after release 0.10.12
```
klmcadams: Is that comment still relevant?

dipinknair: I dont know if it can be removed.

jorgepiloto (approved):
Thanks for fixing this, @dipinknair.

PR #1286, opened 2025-08-21
CHORE: Move `mechanical-env` script to PyMechanical

dipinknair:
If pdm backend works fine, #1285 then `mechanical-env` script will be moved to pymechanical so that we can remove `ansys-mechanical-env` package


dipinknair:
@koubaa Please note that the location of the `mechanical-env` script and the name of the `find-mechanical` file are provisional and may be updated following the review process

pyproject.toml
```diff
@@ -75,29 +75,34 @@ doc = [
     "sphinxcontrib-websupport==2.0.0",
     "ansys-tools-visualization-interface>=0.2.6",
 ]
-# Viz option will be removed in future versions
-viz = [
-    "ansys-tools-visualization-interface>=0.2.6",
-    "usd-core==25.8",
-]
+
```
koubaa, line 78: why are we removing the viz option in this PR?

pyproject.toml
```diff
@@ -27,7 +27,6 @@ classifiers = [
 
 dependencies = [
     "ansys-api-mechanical==0.1.3",
-    "ansys-mechanical-env==0.1.16",
```
koubaa, line 30: what happens if the user has both `ansys-mechanical-env` installed and the newest version of pymechanical?

dipinknair: `mechanical-env` script get replaced from whichever package installed most recently.

PR #1312, opened 2025-09-11
FEAT: Modify installation of python inside container

dipinknair:
Ref : https://github.com/ansys/pymechanical-stubs/pull/125

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -318,79 +318,79 @@ jobs:
           path: mechanical_tests_log-${{ matrix.mechanical-version }}.txt
           retention-days: 7
 
-  embedding-rpc-tests:
-    name: Embedding rpc testing and coverage
-    runs-on: public-ubuntu-latest-8-cores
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_cd.yml
```diff
@@ -159,6 +160,33 @@ jobs:
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Pull, launch, and validate Mechanical service
+        env:
+          MECHANICAL_IMAGE: ${{ matrix.test_container }}
+        run: |
+          docker pull ${{ env.MECHANICAL_IMAGE }}
```
RobPasMue, line 184: There are ways to do this without pulling down the image (which takes time)

See  https://github.com/ansys/pyansys-geometry/blob/5bf5f74e5287882b9f0159084d85f58196846e51/.github/workflows/nightly_docker_test.yml#L59-L86

RobPasMue (approved):

PR #1336, opened 2025-09-29
fix: Fix payload in webhook

klmcadams:
Fix the json payload in the webhook to correctly create messages in the Teams channel. See picture below for successful message posting:

nightlyrunmessage


koubaa (approved):

dipinknair:
You can also take build date and publish that too. It would be nice to see build date of Mechanical.

PR #1351, opened 2025-10-10
FIX: Use different method to check for initialize message

klmcadams:
Use the docker logs command and setting auto flush to 0 to see if "WB Initialize Done" is in the output

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -256,8 +256,6 @@ jobs:
           echo "Sleep for 60 seconds"
           sleep 60  # give time to start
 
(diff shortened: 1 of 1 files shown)


klmcadams:
@koubaa it looks like it still takes about 60 seconds:

[Start time 14:19:22.46801](https://github.com/ansys/pymechanical/actions/runs/18531919434/job/52816842210?pr=1351#step:3:49)
[WB Initialize Done 14:20:06.053049](https://github.com/ansys/pymechanical/actions/runs/18531919434/job/52816842210?pr=1351#step:3:1024)

PR #1357, opened 2025-10-17
CI: Continuously check for WB Initialize Done message

klmcadams:
Continuously check for the "WB Initialize Done" message instead of sleeping for 60 seconds automatically in case it finishes sooner than 60s

Thanks @RobPasMue for the code! He implemented it here: https://github.com/ansys/pyansys-workflows/blob/4aced265e3513063caf72aeec6920a879e169801/.github/workflows/fluent-mechanical.yml#L190-L212

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -271,6 +271,7 @@ jobs:
           done
 
           # Final check
+          docker logs ${{ env.DOCKER_MECH_CONTAINER_NAME }} > log.txt
           if ! grep -q 'WB Initialize Done' log.txt 2>/dev/null; then
             echo "ERROR: Mechanical failed to initialize within ${max_wait} seconds"
             echo "=== Last 50 lines of log.txt ==="


.github/workflows/ci_cd.yml
```diff
@@ -253,11 +253,30 @@ jobs:
...
+              echo "Mechanical initialized successfully after ${elapsed} seconds"
+              break
+            fi
+            echo "Waiting for initialization... (${elapsed}/${max_wait}s)"
+            sleep $check_interval
+            elapsed=$((elapsed + check_interval))
+          done
+
+          # Final check
+          if ! grep -q 'WB Initialize Done' log.txt 2>/dev/null; then
```
RobPasMue:
```suggestion
          # Final check
          docker logs ${{ env.DOCKER_MECH_CONTAINER_NAME }} > log.txt
          if ! grep -q 'WB Initialize Done' log.txt 2>/dev/null; then
```

PR #1377, opened 2025-11-03
CHORE: Update runners

dipinknair:
Updating runners so that it has enough space to pull Mechanical image.


koubaa:
I don't the issue has anything to do with the number of cores, it's probably more related to the amount of disk space and RA. Do the 16 core runners ahve more disk space and ram?

klmcadams (approved):

PR #1390, opened 2025-11-10
FEAT: Embedding with UI

koubaa:
Allow interactive mode with embedding using `interactive=True` kwarg.

This is supported on windows only and only for the dev version of mechanical.

This requires some special handling of the shell since the python main thread and the mechanical main thread can not normally be shareed while the user interface is open.

I have no idea how to test any of this, but I think it is still worth releasing in some kind of experimental mode

### DIFF
--- a/src/ansys/mechanical/core/embedding/ipython_shell.py
+++ b/src/ansys/mechanical/core/embedding/ipython_shell.py
@@ -197,8 +197,8 @@ def post_ipython_blocks():
     EXEC_THREAD.start()
 
     # Patch IPython to delegate to your thread
-    InteractiveShell.run_cell = (
-        _run_cell_in_thread.__get__(InteractiveShell.instance(), InteractiveShell)  # codacy: ignore
+    InteractiveShell.run_cell = _run_cell_in_thread.__get__(
+        InteractiveShell.instance(), InteractiveShell
     )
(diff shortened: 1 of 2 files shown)


src/ansys/mechanical/core/embedding/ipython_shell.py
```diff
@@ -0,0 +1,221 @@
...
+
+def _exec_from_queue(shell) -> bool:
+    """Worker function for ipython execution.
+
+    Return whether to break out of loop
+    """
+    global CODE_QUEUE
+    global ORIGINAL_RUN_CELL
+    global RESULT_QUEUE
+    global SHELL_HOOKS
```
klmcadams: Do you need these global definitions within the functions if they're already set at the top of the file? See this codacy alert: https://app.codacy.com/gh/ansys/pymechanical/pull-requests/1390/issues#issue-d44d5f95cde867c4a050ecc693f6b1d6

Image

src/ansys/mechanical/core/embedding/app.py
klmcadams: Is it possible to add tests (maybe mock tests) for this code?

koubaa: Not at the moment, no.

PR #1424, opened 2025-12-02
FEAT: Add option to chose license check out when starting App

dipinknair:
`start_license` option to chose which license to checkout when starting app


tests/embedding/test_app.py
```diff
@@ -612,6 +612,27 @@ def test_app_start_readonly(run_subprocess, pytestconfig, rootdir, printer):
         assert "The app is started in read-only mode" in stdout
 
 
+@pytest.mark.minimum_version(261)
+@pytest.mark.embedding
+def test_app_start_license(run_subprocess, pytestconfig, rootdir, printer):
+    """Test the app is started with a specific license."""
```
sbradrick: Verify the app...

Test that the app...

koubaa (approved):

PR #1431, opened 2025-12-08
FIX: zizmor errors

dipinknair:
With new version of ansys/action, these error are showing up in CI/CD. https://github.com/ansys/pymechanical/actions/runs/20039616284/job/57470081508

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -51,7 +51,7 @@ updates:
         - "minor"
         - "patch"
     cooldown:
-      default-days: 7
+      default-days: 5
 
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -73,3 +75,8 @@ updates:
           update-types:
           - "minor"
           - "patch"
+    cooldown:
+      default-days: 5
+      semver-major-days: 30
+      semver-minor-days: 7
+      semver-patch-days: 3
```
klmcadams:
```suggestion
```

PR #1437, opened 2025-12-10
chore: replace `ansys-tools-path` with `ansys-tools-common` 

dipinknair:
as title says


pyproject.toml
```diff
@@ -29,7 +29,7 @@ dependencies = [
     "ansys-api-mechanical==0.1.3",
     "ansys-mechanical-stubs==0.1.9",
     "ansys-pythonnet>=3.1.0rc6",
-    "ansys-tools-path>=0.3.1",
+    "ansys-tools-common==0.4.0",
```
RobPasMue:
```suggestion
    "ansys-tools-common>=0.4.0",
```

PR #1566, opened 2026-03-24
ci: skip-doc label to skipping doc build

dipinknair:
This is useful since doc-build takes ~17 mins for every run. For PRs like pre-commit update we dont need to run doc-build or any other similar usecases.

Added case to skip tests also. This is useful for PR with only documentation changes


klmcadams (approved):
This is great, thanks!

klmcadams (approved):
LGTM, thanks!

PR #1569, opened 2026-03-25
feat: Add wbtrace option for logger - embedding

### LINKED ISSUE #355: WBTRACING should be covered in the logging system
low priority, but some errors may occur in the COM implementation of mechanical which traces to the stdout if an env var is set. There should be a way to combine/configure this with PyMechanical

dipinknair:
Partially closes  wbtracing issue
---
``` python
from ansys.mechanical.core.embedding.logger import Configuration
import logging

Configuration.configure(level=logging.ERROR, to_stdout=True, wbtracing=True)
# -- or after the fact --
Configuration.set_wbtracing(True)

app = App(version=271)
print(app)
```


klmcadams (approved):

koubaa:
The changes are fine but I don't think it really closes #355 . For WBTRACING to be fully covered it should write to the same location as the logs (file, stdout, etc).

dipinknair:
@copilot unlink the issue

PR #1601, opened 2026-04-27
fix: awp root error

dipinknair:
from https://github.com/ansys/pymechanical/pull/1592
@Jalmenara

### DIFF
--- a/src/ansys/mechanical/core/embedding/find_mechanical.py
+++ b/src/ansys/mechanical/core/embedding/find_mechanical.py
@@ -102,7 +102,7 @@ def cli(version: int, path: str | None = None) -> tuple[int, str]:
             print(latest_version, str(Path(os.environ[f"AWP_ROOT{latest_version}"]) / "aisol"))
             sys.exit(0)
     else:
-        print("No AWP_ROOT### env. variables found. Falling back to default locations.")        
+        print("No AWP_ROOT### env. variables found. Falling back to default locations.")
 
     # Use ansys-tools-path
     _exe = atp.get_mechanical_path(allow_input=False, version=version)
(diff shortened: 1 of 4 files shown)


src/ansys/mechanical/core/embedding/find_mechanical.py
```diff
@@ -101,10 +101,20 @@ def cli(version: int, path: str | None = None) -> tuple[int, str]:
             latest_version = max(versions_found)
             print(latest_version, str(Path(os.environ[f"AWP_ROOT{latest_version}"]) / "aisol"))
             sys.exit(0)
+    else:
+        print("No AWP_ROOT### env. variables found. Falling back to default locations.")
```
klmcadams: Should we specify what the default location is?

src/ansys/mechanical/core/embedding/initializer.py
```diff
@@ -88,7 +88,13 @@ def _get_latest_default_version() -> int:
     """
     awp_roots = [value for key, value in os.environ.items() if key.startswith("AWP_ROOT")]
     if not awp_roots:
-        raise RuntimeError("No Mechanical installations found.")
+        raise RuntimeError(
+            "No AWP_ROOT### env. variables found.\n"
+            "Is Mechanical installed?\n"
+            "If so, please set manually the environment variable.\n"
```
klmcadams:
```suggestion
            "If so, please set the environment variable manually.\n"
```

koubaa (approved):

src/ansys/mechanical/core/embedding/find_mechanical.py
Jalmenara: Ideally yes, but akaik those are defined externally by "ansys/tools/common/path/path.py". I wouldn't take the risk of specifying them here, as they could change.

`/usr/ansys_inc`, `/ansys_inc` or `/install/ansys_inc`

dipinknair: good idea ! updated

dipinknair: I think these path are not going to change and we can mention them here so user knows where ansys-tools-common is looking. @Jalmenara

Jalmenara (approved):
Thanks for picking this up

dipinknair:
@Jalmenara  Thank you for the contribution 💯

PR #1622, opened 2026-05-15
feat: add initialization flag to manage app state

dipinknair:
as title says


src/ansys/mechanical/core/embedding/app.py
```diff
@@ -69,6 +69,8 @@ def _get_default_addin_configuration() -> AddinConfiguration:
 INSTANCES: list[App] = []
 """List of instances."""
 
+INITIALIZED: bool = False
```
RobPasMue: If you are not expecting users to access this variable themselves, you should consider naming it "__INITIALIZED".

This makes the variable module-private - i.e. it can only be leveraged inside this module. However, users can access it through the ``is_initialized()`` just fine if they want the value.

I would apply this same feedback to the previous variable you have (called INSTANCES)

RobPasMue (commented):
Small comment

src/ansys/mechanical/core/embedding/app.py
dipinknair: Thanks for the feedback. I will update it.

src/ansys/mechanical/core/embedding/app.py
```diff
@@ -66,9 +66,11 @@ def _get_default_addin_configuration() -> AddinConfiguration:
     return configuration
 
 
-INSTANCES: list[App] = []
+_INSTANCES: list[App] = []
 """List of instances."""
 
+_INITIALIZED: bool = False
```
RobPasMue, line 72: Docstring missing

src/ansys/mechanical/core/embedding/app.py
```diff
@@ -296,8 +298,9 @@ def __init__(
         self._app = _start_application(configuration, self._version, db_file, additional_args)
 
         self._disposed = False
-        atexit.register(_atexit_embedded_app, INSTANCES)
-        INSTANCES.append(self)
+        atexit.register(_atexit_embedded_app, _INSTANCES)
+        _INSTANCES.append(self)
+        _INITIALIZED = True
```
RobPasMue: There's something I'm missing here...

Why is instances a list? I'm guessing it's because you are expecting multiple instances...

But if that's the case you should only register the atexit method once, because you are registering it for the module variable INSTANCES.. However, here, you are doing this operation for every app you initialize. Something like this should do the trick
```suggestion
        _INSTANCES.append(self)
        if not _INITIALIZED:
            atexit.register(_atexit_embedded_app, _INSTANCES)
            _INITIALIZED = True
```

RobPasMue: If there's only going to be only 1 instance then you should change the ``_INSTANCES`` object to something else.. not a list

dipinknair: Good question — it's a list because gallery builds can spin up multiple instances. We need to track them all so the atexit handler can clean up properly.

RobPasMue (approved):

PR #1623, opened 2026-05-15
ci: update action versions across workflows

dipinknair:
- Update action 
- update python 3.14 in tests


RobPasMue (approved):
LGTM! Makes it cleaner to follow indeed

PR #1634, opened 2026-05-20
fix: analytics test

dipinknair:
Test started failing in candidate image with a behavior change

Previous file structure
- /path/analyticsxxxxx.json
New file structure
- /path/20260507-144828.310/analyticsxxxxx.json

### DIFF
--- a/tests/embedding/test_analytics.py
+++ b/tests/embedding/test_analytics.py
@@ -31,17 +31,15 @@ import pytest
 
 
 @pytest.mark.embedding_scripts
+@pytest.mark.minimum_version(252)  # Analytics are only captured for 252+
 def test_analytics(rootdir, run_subprocess, pytestconfig, tmp_path: pytest.TempPathFactory):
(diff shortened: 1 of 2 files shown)


tests/embedding/test_analytics.py
```diff
@@ -31,29 +31,23 @@
...
     version = pytestconfig.getoption("ansys_version")
-    # Analytics are only captured for 252+
-    if int(version) < 252:
-        return
-
     embedded_py = Path(rootdir) / "tests" / "scripts" / "run_analytics.py"
 
     analytics_env = os.environ.copy()
-    analytics_env["ANS_ENABLE_DATA_ANALYTICS"] = "1"
+    if version == 252:
```
koubaa, line 41: what about 261??

dipinknair: This env not needed from 261 and onwards for analytics to work. This has been removed as well.

koubaa (approved):
approved. I'm not convinced about 261 but I assume you tested it

PR #1638, opened 2026-05-27
feat: add max-length message display

dipinknair:
- upgrade clr loader
- Add message max- length limit


src/ansys/mechanical/core/embedding/loader.py
```diff
@@ -24,21 +24,19 @@
...
 import os
 from pathlib import Path
-import warnings
 
 
 def __get_mono(assembly_dir, config_dir):
     import clr_loader
 
+    # Explicitly pass libmono to use Ansys's bundled mono rather than any system-installed mono.
     libmono = str(Path(assembly_dir) / "libmonosgen-2.0.so")
```
dipinknair, line 33: @koubaa is this the right change if we are upgrading clr-loader ?

klmcadams (approved):
LGTM!

PR #1646, opened 2026-06-04
fix: remove lock file

### LINKED ISSUE #1645: need a way to unlock a mechdat file while opening

mechdat = r'D:\pump_assembly_static.mechdat'
from ansys.mechanical.core.embedding import add_mechanical_python_libraries
app = App(globals=globals(), version=261, db_file=mechdat )

This works only when mechdat is unlocked.

there is existing API as well, in case we open the blank session first followed by opening mechdat file, but when used as follows:
app.open(db_file=mechdat, remove_lock=True)

Image

_No response_

_No response_

dipinknair:
Fixes #1645 
---
```python

# Before: had to open separately, and lock path was wrong anyway
app = App()
app.open("path/to/locked.mechdb", remove_lock=True)

# After: can pass at construction
app = App(db_file="path/to/locked.mechdb", remove_lock=True)
```


tests/embedding/test_app_save_as.py
```diff
@@ -92,3 +97,52 @@ def test_save_as_without_remove_lock_preserves_lock_file(mock_app, tmp_path):
...
+
+@pytest.mark.embedding
+def test_open_remove_lock_unlinks_lock_file(mock_app, tmp_path):
+    """When ``remove_lock=True`` and ``.mech_lock`` exists, it is removed before Open."""
+    project_file = tmp_path / "project.mechdb"
+    project_file.write_text("", encoding="utf-8")
+    mech_dir = tmp_path / "project_Mech_Files"
+    mech_dir.mkdir()
+    lock_file = mech_dir / ".mech_lock"
+    lock_file.write_text("lock", encoding="utf-8")
```
klmcadams: I think you could make these lines into a function since they repeat in each of the tests. Otherwise, it looks good to me!

klmcadams (approved):

PR #1651, opened 2026-06-09
feat: add doc-deploy-pr label so that it only runs if label is enabled

dipinknair:
as title says

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -155,7 +155,6 @@ jobs:
     if: |
       always() &&
       needs.docs.result == 'success' &&
-      github.event_name == 'pull_request' &&
       contains(github.event.pull_request.labels.*.name, 'deploy-pr-doc')
     permissions:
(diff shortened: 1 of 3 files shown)


klmcadams (approved):

.github/workflows/ci_cd.yml
```diff
@@ -152,7 +152,11 @@ jobs:
     name: "Deploy PR documentation"
     runs-on: ubuntu-latest
     needs: [docs]
-    if: always() && needs.docs.result == 'success'
+    if: |
+      always() &&
+      needs.docs.result == 'success' &&
+      github.event_name == 'pull_request' &&
+      contains(github.event.pull_request.labels.*.name, 'deploy-pr-doc')
```
RobPasMue:
```suggestion
      contains(github.event.pull_request.labels.*.name, 'deploy-pr-doc')
```
I believe the first condition is redundant. The labels are only available on PRs so it is assumed that if the contains returns true, you are on a PR

RobPasMue (approved):

